PR #8, opened 2016-02-21
Added EER Diagram and updated README with more information.

Raistlfiren:
Added more explicit information in README and added PNG of EER Diagram.


README.md
```diff
@@ -6,14 +6,39 @@ Contains a dockerfile to run our database in a contained environment, along with
...
+    docker exec -i faf-db mysql -uroot -p<wantedpassword> <db_name> < db-structure.sql
+
+Import Data
+
+    docker exec -i faf-db mysql -uroot -p faf_db  < db-data.sql
+
+Now the database should be ready to go!
+
+#EER Diagram
+![EER Diagram of FAF DB](https://raw.githubusercontent.com/FAForever/db/master/documentation/EER_Diagram.png)
```
Sheeo: While it's nice to have an EER diagram, it would be much nicer if the database was actually meaningful with such a display. Currently, this isn't the case.

The structure file should see some sanitization and reach a more stable state before we publish something like this, imo. It will just require too many updates (Besides; we can generate these locally any time -- so I question the usefulness of having it in this repository).

README.md
```diff
@@ -6,14 +6,39 @@ Contains a dockerfile to run our database in a contained environment, along with
...
+    netstat -atn | grep 3306
+
+Find containers IP (Container ID can be found under docker ps)
+
+    docker inspect <container_id> (IP is under IPAddress in NetworkSettings)
 
-    docker run -d --name faf-db -e MYSQL_ROOT_PASSWORD=<wanted_password> faf-db
+Connecting to the mysql instance
+
+    mysql -uroot -p -h <ip_address>
```
Sheeo: Why guide users on how to connect from the host system if we just below run imports using the container?

README.md
```diff
@@ -6,14 +6,39 @@ Contains a dockerfile to run our database in a contained environment, along with
...
 
+Quick overview of Docker can be found:
+[Docker Quick Start Guide](https://docs.docker.com/engine/quickstart/)
+
 Build the container using
 
     docker build -t faf-db .
 
-Run using
+Run using(This will publish on port 3306 of the container)
```
Sheeo: This is a bit misleading.

It will actually publish on port 3306 of the _host system_. If you are running OS X / Windows, the host system happens to be a virtual machine that docker creates for you, but otherwise if you run linux, it is your actual host OS exposing this port.

README.md
Raistlfiren: Yeah, I follow you. It is just sometimes easier to understand how the database is laid out with a diagram. You can easily generate the diagram, but it requires a tool like phpmyadmin or MYSQL workbench. My thoughts are to make it as easy as possible to understand how things work. Removing it isn't a big deal though.

README.md
Raistlfiren: It is elementary, but it might be useful for someone trying to understand how to connect to the database.

README.md
Raistlfiren: Oh interesting, I misunderstood that. I will change that information.

Brutus5000:
:+1: Would have saved me 2 hours of my sunday learning about docker...

README.md
IDragonfire: Maybe at this not to a repo, but to a wiki?
And in the repository add scripts to generate a diagram ...

PR #10, opened 2016-02-28
DB Structure import fix

Raistlfiren:
Removed comma from last statement in sql, breaking import of db structure.


Sheeo:
Thanks, I missed this in a previous change. Though we need to restructure this, so I'm reverting that change for now instead.

PR #11, opened 2016-05-02
Updated readme instructions so that they work out of the box.

Brutus5000:
For developers who haven't worked with docker (like me) the instructions should work out of the box. So I digged into it and added the missing statements.

If have further added the hint for the MySQL version, because I tried to host the database on my Synology NAS, which has a very old MariaDB version. On old versions, you may run into the error 
> #1293 - Incorrect table definition; there can be only one TIMESTAMP column with CURRENT_TIMESTAMP in DEFAULT or ON UPDATE clause

on importing data-structure.sql.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Contains a dockerfile to run our database in a contained environment, along with
 
 Get [docker](http://docker.com).
 
-Build the container using
+Build the container from the FAForever/db-directory using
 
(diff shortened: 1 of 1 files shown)


Brutus5000:
Oh I just saw there is already a pull request from Raistlfiren. Therefore I close this.

PR #19, opened 2016-05-18
Add table `teamkills` for server-issue #95

Brutus5000:
Prerequisite to work on https://github.com/FAForever/server/issues/95
Table structure is based upon the new GPGNet-command from https://github.com/FAForever/fa/pull/903

### DIFF
--- a/migrations/add_teamkills-ab3d56fe949550d2ac113a148582e4c7c690e25c.sql
+++ b/migrations/add_teamkills-ab3d56fe949550d2ac113a148582e4c7c690e25c.sql
@@ -1,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS `teamkills` (
+  `teamkiller` mediumint(8) unsigned NOT NULL COMMENT 'login of the player who performed the teamkill',
+  `victim` mediumint(8) unsigned NOT NULL COMMENT 'login of the player who got teamkilled and reported the tk',
+  `game_id` int(10) unsigned NOT NULL COMMENT 'game-id where teamkill was performed',
+  `gametime` mediumint(8) unsigned NOT NULL COMMENT 'time of game in seconds when tk was performed',
+  `reported_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  INDEX (game_id));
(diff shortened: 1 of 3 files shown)


Crotalus:
Maybe add a reference to what game it happened as well if someone wants to investigate it more carefully

Brutus5000:
Good idea, done.

Brutus5000:
There was a slight misunderstanding with the GPGNet-Command on my side. The killtime is the time in seconds of game running and not the local time, therefore we should save this in the db as well.

migrations/add_teamkills-ab3d56fe949550d2ac113a148582e4c7c690e25c.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS `teamkills` (
+  `teamkiller` mediumint(8) unsigned NOT NULL COMMENT 'login of the player who performed the teamkill',
+  `victim` mediumint(8) unsigned NOT NULL COMMENT 'login of the player who got teamkilled and reported the tk',
+  `game` int(10) unsigned NOT NULL COMMENT 'game-id where teamkill was performed',
+  `gametime` mediumint(8) unsigned NOT NULL COMMENT 'time of game in seconds when tk was performed',
+  `reported_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP);
```
Sheeo: I'm being nitpicky here (It's a rather small problem), but please configure your editor to insert trailing newlines at EOF.

This matters for certain CLI tools :)

Sheeo:
It's good for a first -- I suggest we add an index on `game` (Shouldn't we call this column `game_id`?) immediately though, since that's probably what we're going to be querying on.

Also, since I just merged Dragonfire's autogenerated db-structure.sql file, would you mind rebasing this onto the latest develop? You'll want to run Dragonfire's regenerate script.

Brutus5000:
Rebased and all suggestions implemented.

Sheeo:
👍

Brutus5000:
Connects to #95 (waffle.io reference)

Issue #15, opened 2016-05-16
db-structure.sql autogenerate


IDragonfire:
db-structure.sql can be automatic generated in the builld process ...

Sheeo:
Does this mean we want to? I wouldn't want travis committing back into this repo and we do want a plain .sql file that's easily runnable to construct the database structure.

IDragonfire:
The current sql file is mostly automatic generated from mysqld...  What is a plain sql file? Ruby on rails for example has such a Schema.sql ...very useful to see of migration working over time... And it is just another task to do by Hand and to check....

IDragonfire:
btw. it must not generated by travis and commited ... can be done by dockerfile at runtime ... or a build script ... I will provide a readme or shell script for it ...

PR #20, opened 2016-05-18
Add script to auto generate db-structure.sql #15

IDragonfire:
Regenerate db-structure.sql


create-db-structure.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+rm db-structure.sql
+docker exec faf-db mysqldump -pbanana --no-data --skip-add-drop-table faf_test > db-structure.sql
```
Sheeo: Of course, this assumes `banana` as password. Perhaps we should take it on commandline/stdin?

Sheeo:
👍 thanks :)

IDragonfire:
thanks for merging ;)

PR #21, opened 2016-06-15
Use DEFAULT CURRENT_TIMESTAMP and remove triggers

micheljung:
Now possible as we have new mysql version everywhere


micheljung:
Tested both, `db-data.sql` and migration script. Fixed some mistakes, looks good now.

PR #24, opened 2016-06-18
Clean up map tables

micheljung:
* Split `table_map` into `map` and `map_version`
* Merge `table_map_unranked` into `map.ranked`
* Merge `table_map_uploaders` into `map.uploader`
* Create view `table_map` for backwards compatibility
* Add `test_migration.sh` for easy (currently manual) testing of new SQL scripts

### DIFF
--- a/db-structure.sql
+++ b/db-structure.sql
@@ -1157,30 +1157,62 @@ CREATE TABLE IF NOT EXISTS `swiss_tournaments_players` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 
 --
--- Table structure for table `table_map`
+-- Table structure for table `map`
 --
 
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
-CREATE TABLE IF NOT EXISTS `table_map` (
+CREATE TABLE IF NOT EXISTS `map` (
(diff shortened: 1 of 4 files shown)


db-structure.sql
```diff
@@ -1157,29 +1157,40 @@ CREATE TABLE IF NOT EXISTS `swiss_tournaments_players` (
...
-  `mapuid` mediumint(8) unsigned NOT NULL,
-  PRIMARY KEY (`id`),
-  UNIQUE KEY `Combo` (`name`,`version`),
-  UNIQUE KEY `map_filename` (`filename`),
-  KEY `mapuid` (`mapuid`)
-) ENGINE=InnoDB AUTO_INCREMENT=5692 DEFAULT CHARSET=latin1;
+  `max_players` decimal(2,0) NOT NULL,
+  `size_x` decimal(4,0) NOT NULL,
+  `size_y` decimal(4,0) NOT NULL,
+  `version` decimal(4,0) NOT NULL UNIQUE,
```
Sheeo: This shouldn't be `UNIQUE`. Rather you want a composite `(map_id,version)` key with a uniqueness constraint. This is what causes travis to fail.

micheljung: Yes, thanks!

PR #27, opened 2016-07-11
Add more null constraints to lobby_ban

ChrisKitching:
Represent infinite bans as 200-year bans instead. The new default
for this field is this, to retain compatibility with code that
doesn't specify the field.

Prune bans with null user id, because WTF.

### DIFF
--- a/migrations/nonnull_bans-2ee16b581c35a43bfb5e831e42728bb66dea5ce5.sql
+++ b/migrations/nonnull_bans-2ee16b581c35a43bfb5e831e42728bb66dea5ce5.sql
@@ -1,0 +1,9 @@
+-- These are garbage
+DELETE FROM lobby_ban WHERE idUser IS NULL;
+
+-- Represent infinite bans with large numbers, not NULL.
+UPDATE lobby_ban SET expires_at = '2222-07-21' WHERE expires_at IS NULL;
+
+ALTER TABLE lobby_ban 
+    MODIFY expires_at datetime NOT NULL DEFAULT '2222-07-21',
+    MODIFY idUser mediumint(8) unsigned NOT NULL;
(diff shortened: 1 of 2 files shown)


migrations/nonnull_bans-2ee16b581c35a43bfb5e831e42728bb66dea5ce5.sql
```diff
@@ -0,0 +1,9 @@
+-- These are garbage
+DELETE FROM lobby_ban WHERE idUser IS NULL;
```
Sheeo: These also don't exist, so that's something ;)

migrations/nonnull_bans-2ee16b581c35a43bfb5e831e42728bb66dea5ce5.sql
```diff
@@ -0,0 +1,9 @@
...
+
+-- Represent infinite bans with large numbers, not NULL.
+UPDATE lobby_ban SET expires_at = '2222-07-21' WHERE expires_at IS NULL;
+
+ALTER TABLE lobby_ban 
+    MODIFY expires_at datetime NOT NULL DEFAULT '2222-07-21',
```
micheljung: ` COMMENT 'When the ban expires',`

Sheeo: Do we really need that comment when the key name is so descriptive and it's no longer nullable?

ChrisKitching: Not at all. Comments are rarely necessary (database schema are generally fairly self-explanatory, really). Stuff so obvious is not worth the spam :P

micheljung: Well then remove it in the db-structure.sql, too, to keep it consistent ;-)

PR #30, opened 2016-08-10
add sql alchemy map tables

Cheyans:


micheljung:
If those are generated, then we shouldn't even check them in but instead have them generated automatically on the dev's machine :-)

Cheyans:
Given that the code is used by other repositories I don't think that really makes sense. If it was documentation or something like that sure, but it's an API for other repos.

PR #31, opened 2016-09-07
Add a column for clan tag colours in faf db

HardlySoftly:


HardlySoftly:
See https://github.com/FAForever/db/pull/31 and https://github.com/FAForever/api/pull/101

db-structure.sql
```diff
@@ -307,6 +307,7 @@ CREATE TABLE `clan_list` (
   `clan_tag` varchar(3) DEFAULT NULL,
   `clan_founder_id` mediumint(8) DEFAULT NULL,
   `clan_leader_id` mediumint(8) DEFAULT NULL,
+  `clan_tag_color` varchar(30) DEFAULT NULL,
```
micheljung: Adding this column here is not enough, you need to write a migration file that adds this column to an existing database.

Sheeo:
It was decided we're not doing this on a previous dev meeting.

If you have a full implementation please reopen and link corresponding pr's on the server and the client(s).

HardlySoftly:
I wasn't in any previous dev meetings, and don't recall agreeing to not implement the clan tag colour feature.

I am aiming for an implementation which gives full API support, once this is ready I will re-implement the client side.

HardlySoftly:
On the point of consistency of the word "color/colour", it was my understanding that the codebase was to be in American English, in which case the appropriate spelling to use would be "color" (this spelling is also already in use within that file - https://github.com/HardlySoftly/db/blob/7a71f16af191f75035064e25f880527f5e4a658f/db-structure.sql#L480 )

Since the branch name isn't part of what I consider the code, I instead decided to use the correct (:P) spelling of colour.

HardlySoftly:
https://github.com/FAForever/db/pull/32 is the new PR

micheljung:
It was decided that nobdy *wants* to do it, not that we're against having this feature.
Now somebody wants to do it, so let him do it.

PR #32, opened 2016-09-08
Add clan tag colour support

HardlySoftly:
Now including migration support

### DIFF
--- a/migrations/add_clan_tag_colors-150bb43b63f44549c415ae56853632f8779f8cd4.sql
+++ b/migrations/add_clan_tag_colors-150bb43b63f44549c415ae56853632f8779f8cd4.sql
@@ -1,0 +1,1 @@
+ALTER TABLE clan_list ADD COLUMN `clan_tag_color` varchar(30) DEFAULT NULL COMMENT 'Color of the clan''s tag in chat';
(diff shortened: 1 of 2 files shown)


migrations/add_clan_tag_colors-150bb43b63f44549c415ae56853632f8779f8cd4.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE clan_list ADD COLUMN `clan_tag_color` varchar(30) DEFAULT NULL COMMENT 'Color of the clan''s tag in chat';
```
micheljung: Either add `COMMENT` in the db-structure as well, or remove it here... consistency :-)

micheljung: And good question from Brutus, why `varchar(30)` and not `varchar(6)` (hex code)?

Brutus5000:
VARCHAR(30) can be anything like "red-like-my-enemies-blood"
I propose VARCHAR(6) for a RGB-Hexcode.

HardlySoftly:
I was thinking along the lines of human readable colour names http://htmlcolorcodes.com/color-names/

PR #45, opened 2016-12-22
Feature/surveys

Cheyans:


migrations/V10__surveys.sql
```diff
@@ -0,0 +1,44 @@
+CREATE TABLE surveys (
+  id INTEGER NOT NULL AUTO_INCREMENT,
+  name VARCHAR(256) NOT NULL,
+  author_id MEDIUMINT(8) unsigned NOT NULL,
+  start_dtm TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  end_dtm TIMESTAMP NOT NULL DEFAULT '2038-01-18 03:14:07',
```
Brutus5000: I see the advantage of not having to deal with null values seperately, but do you really want to see "this survey ends 2038" instead of "this survey has no end date yet"?

Cheyans: It allows for the constraint to work. A null value would break it.

micheljung: Then remove the constraint

migrations/V10__surveys.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE surveys (
+  id INTEGER NOT NULL AUTO_INCREMENT,
+  name VARCHAR(256) NOT NULL,
```
micheljung: This should be 255, 256 uses 1 more byte for no gain

migrations/V10__surveys.sql
```diff
@@ -0,0 +1,43 @@
...
+  author_id MEDIUMINT(8) unsigned NOT NULL,
+  start_dtm TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  end_dtm TIMESTAMP NULL,
+  update_dtm TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
micheljung: Please keep naming consistent: create_time and update_time

migrations/V10__surveys.sql
```diff
@@ -0,0 +1,43 @@
...
+  PRIMARY KEY(id),
+  FOREIGN KEY (author_id) REFERENCES login (id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE TABLE survey_questions (
+  id INTEGER NOT NULL AUTO_INCREMENT,
+  survey_id INTEGER NOT NULL,
+  question TEXT CHARACTER SET utf8 NOT NULL,
+  multi_answer TINYINT NOT NULL DEFAULT FALSE COMMENT 'If user can select multiple options',
+  display_order SMALLINT NULL COMMENT 'Relative to all questions found under one survey_id',
```
micheljung: Please add create_time and update_time

migrations/V10__surveys.sql
```diff
@@ -0,0 +1,43 @@
...
+  PRIMARY KEY (id),
+  FOREIGN KEY (survey_id) REFERENCES surveys (id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE TABLE survey_question_options (
+  id INTEGER NOT NULL AUTO_INCREMENT,
+  question_id INTEGER NOT NULL,
+  `option` TEXT CHARACTER SET utf8 NOT NULL,
+  open_response TINYINT NOT NULL DEFAULT FALSE COMMENT 'If the answer allows for custom user input, user reponse should be put in corresponding survey_ansers.response column',
+  display_order SMALLINT NULL COMMENT 'Relative to all questions found under one qustion_id',
```
micheljung: Please add create_time and update_time

migrations/V10__surveys.sql
```diff
@@ -0,0 +1,43 @@
...
+  PRIMARY KEY (id),
+  FOREIGN KEY (question_id) REFERENCES survey_questions (id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE TABLE survey_answers (
+  id INTEGER NOT NULL AUTO_INCREMENT,
+  user_id MEDIUMINT(8) unsigned NOT NULL,
+  option_id INTEGER NOT NULL,
+  response TEXT CHARACTER SET utf8 NULL COMMENT 'Should be populated only if corresponding survey_question_option.open_response column is true',
+  answer_dtm TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
micheljung: why call it `dtm` if it's not `DATETIME` but `TIMESTAMP`?

migrations/V10__surveys.sql
Cheyans: added

migrations/V10__surveys.sql
Cheyans: added

PR #75, opened 2017-02-15
Fix #74 improve dockerization

### LINKED ISSUE #74: Improve Dockerization
IMO this project needs the following improvements:

- Initialize the database automatically during container creation *if the database does not yet exist* instead of having the user execute a setup script manually. Also it should be "production ready", meaning that it assumes to be a productive DB and therefore does not load any test data into the database.
- In order to integrate well with `faf-stack`, this project should also create databases and users for other tools like IRC, Mediawiki and so on since it's the only project that should have DB root permissions.
- Since `faf-server` (and potentially `faf-api` or others) doesn't work properly without some standard "featured" mods in the database, such core data should be added during the DB initialization (it could be added by faf-server if it's missing, but that's not smart since other projects may rely on it as well)
- The Dockerfile should specify a healthcheck (new Docker feature)
- The migration script should be part of the container and executed within the container, not from outside

micheljung:
See #74


init/0_init_db_and_users.sh
```diff
@@ -0,0 +1,27 @@
...
+create "faf" "faf-api" "${MYSQL_API_PASSWORD}"
+create "faf" "faf-legacy-apps" "${MYSQL_LEGACY_APPS_PASSWORD}"
+create "faf" "faf-legacy-live-replay-server" "${MYSQL_LEGACY_LIVE_REPLAY_SERVER_PASSWORD}"
+create "faf" "faf-legacy-secondary-server" "${MYSQL_LEGACY_SECONDARY_SERVER_PASSWORD}"
+create "faf" "faf-legacy-updater" "${MYSQL_LEGACY_UPDATER_PASSWORD}"
+create "faf-murmur" "faf-murmur" "${MYSQL_MURMUR_PASSWORD}"
+create "faf" "faf-server" "${MYSQL_SERVER_PASSWORD}"
+create "faf-softvote" "faf-softvote" "${MYSQL_SOFTVOTE_PASSWORD}"
+create "faf-anope" "faf-anope" "${MYSQL_ANOPE_PASSWORD}"
+create "faf-unrealircd" "faf-unrealircd" "${MYSQL_UNREALIRCD_PASSWORD}"
```
duk3luk3: unrealircd doesn't need a MYSQL DB

micheljung: Well spotted! thx

Issue #67, opened 2017-01-30
Add Permission System for Mod App


IDragonfire:
Requirement from the Mod App Specification

```
Permission System
Not every mod can use every feature
```

I suggest to introduce a classic permission system:
* You ask for a special permission, e.g. `ban.create`, `ban.delete`
* Admins can define roles with a set of permissions, e.g.role `mod` with permission `ban.*, avatar.*`
* Admins can assign permissions or roles to a user

# Step 1
* We introduce Permissions

# Step 2
* We introduce Roles

micheljung:
see also: http://www.erdiagrams.com/datamodel-rbac-idef1x.html

Just a 1min Google. I'm pretty sure we can find existing SQLs or so for some sort of "proven model"

IDragonfire:
http://slackspace.de/articles/roles-permissions-with-spring-security-3/

IDragonfire:
https://github.com/FAForever/db/pull/76

Issue #51, opened 2016-12-30
Remove obsolete tables


micheljung:
There are a few tables that are no longer used (or have never been used) and should therefore be removed:

* `auth_group` (never used)
* `auth_group_permissions` (never used)
* `auth_permission` (never used)
* `auth_user` (never used)
* `avatars_list_copy_812015` (unused, empty)
* `bet` (no longer used, data has no value for us. Some people were able to "cheat".)
* `db_version` (unused, was part of the old migration script)
* `game_min_rating` (unused, redundant)
* `game_replays` (only used by non-working, unused command `command_list` in secondaryServer)
* `game_replays_old` (unused, empty)
* `ladder_season_3_safe` (unused, empty)
* `ladder_divisions` (unused, useless data, superseded by `ladder_division`)
* `login_with_duplicated_users` (unused, empty)
* `replay_vault` (unused, superseded by `game_stats`
* `smurf_table` (unused, obsolete)
* `table_map_broken` (unused, superseded by "hidden" flag in new table structure)
* `test_game_replays` (unused)
* `tmp` (unused)
* `updates` (obsolete, unused)
* `vault_admin` (unused, obsolete data (2 entries))
* `view_global_rating` (unused, isn't even a view)

IDragonfire:
What is with the two remaining auth tables?
* `auth_user_user_permissions` https://github.com/search?q=org%3AFAForever+auth_user_user_permissions&type=Code
* `auth_user_user_permissions` https://github.com/search?utf8=%E2%9C%93&q=org%3AFAForever+auth_user_user_permissions&type=Code

IDragonfire:
Do we really need a table for `faction` instead of an enum?
![image](https://cloud.githubusercontent.com/assets/1217999/25969408/0fbd7a04-3695-11e7-8765-e2c549d40f13.png)

IDragonfire:
What is with the table `steam_link_request`?
Also empty and not used ...
https://github.com/FAForever/db/search?utf8=%E2%9C%93&q=steam_link_request&type=`

micheljung:
Please create a new issue for additional tables. This one belongs to a milestone that has already been closed.

PR #76, opened 2017-03-06
[2] Add Permission system #67, e.g. for modapp

IDragonfire:
### Please consider the affect on the following components
- [x] python-server: currently using `lobby_admin`, view created, `auth_login` for simple level requests 
- [x] python-api: no need for permission
- [x] java-api: waiting on permission system, currently modapp is blocked by this PR
- [x] website: no need for permission

I chose the prefix `login_` to keep the table together and avoid confusion with the `auth_` tables ...


migrations/V21__permission-system.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE `login_permission` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT COMMENT '',
+  `name` VARCHAR(40) NOT NULL COMMENT 'name of the permission',
+  PRIMARY KEY (`id`)  COMMENT '')
+COMMENT = 'Manage all available permissions';
```
micheljung: Foreign key to `login` is missing, and I'd remove all empty comments and the ones that don't add add additional information (`login_role.name` is the "name of the role" /giphy you don't say) :D

IDragonfire: NO !!
This is only a list of all permissions

IDragonfire:
image

micheljung:
Good point, please add `create_time` and `update_time` to all tables as well

Brutus5000:
Please add constraints.
- a `login_permission` should not be removable as long as it is referenced by  a `login_role_permission`
- a `login_role_permission` should not be removable as long as it is referenced by  a `login_role`
- a `login_role` should not be removable as long as it is referenced by  a `login_user_role`

Otherwise we'll end up in a mess sooner or later.

IDragonfire:
> a login_permission should not be removable as long as it is referenced by a login_role_permission

https://github.com/FAForever/db/pull/76/files#diff-bbf50ddfae31b48363836194468eb748R25

> a login_role_permission should not be removable as long as it is referenced by a login_role

https://github.com/FAForever/db/pull/76/files#diff-bbf50ddfae31b48363836194468eb748R24

> a login_role should not be removable as long as it is referenced by a login_user_role_assignment

https://github.com/FAForever/db/pull/76/files#diff-bbf50ddfae31b48363836194468eb748R36

This foreign keys shoul be enough, or did I miss something?

Brutus5000:
Right you are, I was missing the `CONSTRAINT` keyword, but it's optional.

duk3luk3:
what are you going to do with the lobby_admin table?

IDragonfire:
It will take a while until the mod app is in a stable state and until this point both will coexist ...

micheljung:
I'd like to name the tables properly from the beginning, otherwise we'll have yet another mess. I don't know why I didn't remove the remaining two auth_ tables as part of #51. They are empty on production, but maybe the legacy map vault is still using it? We have to check. Otherwise we can remove/replace those as part of this PR.

IDragonfire:
@micheljung What is your naming suggestion?

micheljung:
Personally, I like to name things just as they are (so no prefix), but if people here prefer a prefix then I'd go with `auth_` (`login_permission` is pretty confusing`

IDragonfire:
@Brutus5000 @duk3luk3 @Sheeo `auth_` as prefix?  or no prefix? Anything else missing, open?

duk3luk3:
We discussed this in last night's dev conference.

We strongly came down on the side that there is no need for a role/permission "matrix" based model and the drawbacks that come with it.

Our main arguments against the fully normalized role/permission model are:
* Requires clients to hardcode the names of permissions so they can check for them
* Requires more code to determine the permissions of a user from the db, and from the api if REST standards are strictly followed (i.e. no endpoint that performs the joins needed so it can return the set of permissions for a user)

Instead, we want to keep the current "level" based model, with four roles that have a numeric power level, e.g.:

* Level 0 = User
* Level 10 = Tournament Director
* Level 50 = Moderator
* Level 100 = Administrator

So, any user with a power level >= 10 can start and manage tournaments, and any user with a power level >= 50 can kick users from games.

The numeric progression leaves plenty of room for inserting additional levels, without requiring changes to any clients who do not implement the new features coming with the additional levels.

The `lobby_admin` table should be refactored to accomplish this.

micheljung:
But let's rename `lobby_admin` anyway, e.g. to `user`_roles`?

duk3luk3:
Yes. (that's why I said refactor lobby_admin? ;-) )

IDragonfire:
Permission system rejected: https://www.youtube.com/watch?v=IHE1sb8MJ6c&feature=youtu.be#t=1h6m18s

micheljung:
It wasn't rejected. It had been discussed and we came to the conclusion that it doesn't need to be as comprehensive for our current and foreseeable needs (even though I, personally, would build it this way anyway).
If that's wrong, bring it up and we'll come up with a new decision. Just don't be a crybaby, please.

bukajsytlos:
If I may add my two cents, as I wrote during dev conf, I dont like idea of level permissions. Seems to restrictive with very little benefits.

Brutus5000:
The question whether we need "building roles" from permission vs. "just give permission to users" is still a valid point. In an ERP system you have thousands of permissions and want to organize them in roles. But this probably does not apply to FAF.

A simpler approach would be giving just permissions to user. So it's still a single database select for a dedicated permission.

IDragonfire:
> A simpler approach would be giving just permissions to user. So it's still a single database select for a dedicated permission.

Correct, but this is also more work in maintaince, because if you introduce a new module, e.g.  skins ... you most give the permission to every admin and mod ...

IDragonfire:
Implemented `level` system as view, if we have performance problems we can materialized the table ....

duk3luk3:
I'm sorry, but that sounds like the worst possible thing to do.

* Views are slow in mysql - not that that really matters for this case, the permission system shouldn't be under much pressure.
* MySQL doesn't *have* materialized views, so we would need to create a stored procedure to "materialize" the relation between roles and levels.
* There's absolutely no technical reason to do this.

Just like I said before, all that's needed to do is to refactor the `lobby_admin` table.

IDragonfire:
* We have ONE system, the level/rank is a property of the role 
* I know the  MySQL view limitation and if this is an issue the permission api can still manual materialize it, but currently I see not the need 
* I see the technical reason, the server wants a simple fast query for permission and here you have a table/view/api for it on SQL level...

duk3luk3:
> the permission api can still manual materialize it

That would put responsibility for DB consistency outside of the DB. That's a bad idea. You can see why that's a bad idea from the current state of our DB.

IDragonfire:
That is correct, but to increase performance you have either to simply things or introduce redundant data/ cache that introduce other problems you mention...

duk3luk3:
There's no performance problem here in the first place...

*If* we add a role-based permission system, which I'm still against, the view is probably a good idea for backwards compatibility, but then you'll have to call it `lobby_admin` and come up with a backward-compatible way of mapping "moderator" level to 1 and "admin" level to 2 (probably by giving moderators level 100 and admins level 200 and dividing by 100 in the view).

IDragonfire:
Add Permission system #67, e.g. for modapp

IDragonfire:
# clanapp
New use came is the clan app... 
You want to give some members some time powers, e. G. 
* invite new member 
* edit clan data 
*... 

It is possible to use a custom  level system or a role systems ;)

IDragonfire:
> If we add a role-based permission system, which I'm still against, the view is probably a good idea for backwards compatibility, but then you'll have to call it lobby_admin and come up with a backward-compatible way of mapping "moderator" level to 1 and "admin" level to 2 (probably by giving moderators level 100 and admins level 200 and dividing by 100 in the view).

How use `lobby_admin`? On the test server I only see `group` values of `2` ....

IDragonfire:
# Summary
* We have now a `lobby_admin` view for backwards compatiblity
* We have now a `auth_login` view to allow simple server queries

Anything missing?

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
+  FOREIGN KEY (`user_id`) REFERENCES `login`(`id`),
+  FOREIGN KEY (`role_id`) REFERENCES `auth_role`(`id`))
+COMMENT = 'Assign users (login) to a role (a bunch of permissions)';
+
+CREATE 
+    ALGORITHM = UNDEFINED 
+    DEFINER = `root`@`%`
```
micheljung: > The DEFINER and SQL SECURITY clauses specify the security context to be used when checking access privileges at view invocation time

I don't think that should be `root`? Also, `CREATE VIEW` should be sufficient.

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+    SELECT 
+        `login`.`id` AS `login_id`,
+        MAX(`auth_role`.`level`) AS `auth_level`
+    FROM
+        ((`login`
+        LEFT JOIN `auth_user_role_assignment` ON ((`login`.`id` = `auth_user_role_assignment`.`user_id`)))
+        LEFT JOIN `auth_role` ON ((`auth_user_role_assignment`.`role_id` = `auth_role`.`id`)))
+    GROUP BY `login`.`id`;
+
+# Backwords compatible to old `lobby_admin` table
```
micheljung: "Backwards"

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+    (2, 'Administrator', 999);
+
+INSERT INTO `auth_user_role_assignment` (`user_id`, `role_id`) 
+SELECT `user_id`, `group` FROM `lobby_admin`;
+
+DROP TABLE `lobby_admin`;
+
+CREATE 
+    ALGORITHM = UNDEFINED 
+    DEFINER = `root`@`%`
```
micheljung: > The DEFINER and SQL SECURITY clauses specify the security context to be used when checking access privileges at view invocation time

I don't think that should be `root`? Also, `CREATE VIEW` should be sufficient.

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+    DEFINER = `root`@`%` 
+    SQL SECURITY DEFINER
+VIEW `lobby_admin` AS
+    SELECT 
+        `auth_login`.`login_id` AS `user_id`,
+        IF((`auth_login`.`auth_level` >= 999),
+            2,
+            IF((`auth_login`.`auth_level` >= 500),
+                1,
+                0)) AS `group`
```
micheljung: If you want it to be more readable and thus easier to understand, you could use CASE:

```
case 
       when `auth_login`.`auth_level` >= 999 then 2
       when `auth_login`.`auth_level` >= 500 then 1 
       else 0 
end AS `group`
```

Also, CASE is standard SQL and works in most databases. `IF()` is not standard SQL and will not work in other databases, like PostgreSQL.

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+  `name` VARCHAR(40) NOT NULL COMMENT 'e.g. gameban.create or gameban.delete',
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`) )
+COMMENT = 'List of all permissions, assigned over role assignments (auth_role_permission_assignments)';
+
+CREATE TABLE `auth_role` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `name` VARCHAR(40) NOT NULL COMMENT 'e.g. Admin,Mod',
+  `level` MEDIUMINT(4) UNSIGNED COMMENt 'level of this role for greater less comparison in server: Admin >= 999; Mod >= 500;',
```
micheljung: typo `COMMENT`

Sentences start uppercase: `Level of...`

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+CREATE TABLE `auth_role_permission_assignments` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `role_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `permission_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
+  FOREIGN KEY (`role_id`) REFERENCES `auth_role`(`id`),
+  FOREIGN KEY (`permission_id`) REFERENCES `auth_permission`(`id`))
+COMMENT = 'Connect roles (auth_role) and permissions (auth_role_permission_assignments)';
```
micheljung: typo `Connects`

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+COMMENT = 'List of all permissions, assigned over role assignments (auth_role_permission_assignments)';
+
+CREATE TABLE `auth_role` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `name` VARCHAR(40) NOT NULL COMMENT 'e.g. Admin,Mod',
+  `level` MEDIUMINT(4) UNSIGNED COMMENt 'level of this role for greater less comparison in server: Admin >= 999; Mod >= 500;',
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`) )
+COMMENT = 'List of all roles. A bunch of permission forms a role (see auth_role_permission_assignments) and are assigned over auth_user_role_assignment to a user';
```
micheljung: grammar `a bunch of permissions form a role`

migrations/V35__permission-system.sql
```diff
@@ -0,0 +1,80 @@
...
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `name` VARCHAR(40) NOT NULL COMMENT 'e.g. gameban.create or gameban.delete',
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`) )
+COMMENT = 'List of all permissions, assigned over role assignments (auth_role_permission_assignments)';
+
+CREATE TABLE `auth_role` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `name` VARCHAR(40) NOT NULL COMMENT 'e.g. Admin,Mod',
```
micheljung: Typo `Admin, Mod`

migrations/V35__permission-system.sql
IDragonfire: good to know, thanks ;)

migrations/V36__permission-system.sql
```diff
@@ -0,0 +1,72 @@
...
+
+DROP TABLE `lobby_admin`;
+
+CREATE VIEW `lobby_admin` AS
+    SELECT 
+        `auth_login`.`login_id` AS `user_id`,
+        case 
+          when `auth_login`.`auth_level` >= 999 then 2
+          when `auth_login`.`auth_level` >= 500 then 1 
+          else 0
```
micheljung: Shouldn't this view be based on `auth_role`?

IDragonfire: `lobby_admin` is only for backwards compatible and the relevant value is not the `name`, it is the `level` power from my view. In the old model we had only `USER, MOD, ADMIN` maybe in the future we have multiple Mods and then we can handle this over the `level`.

Brutus5000:
The python api **is** affected by this change. See implementation [here](https://github.com/FAForever/api/blob/develop/api/user.py). It is used for avatar management (Jeremy app).

Brutus5000:
Closed in favor of #155

PR #77, opened 2017-03-08
Rework ban

IDragonfire:
* Distinguish between permanent and expireable ban
* Possible to revert/disable ban (but keep the ban in the database)

### DIFF
--- a/migrations/V30__ban-system-ng.sql
+++ b/migrations/V30__ban-system-ng.sql
@@ -1,0 +1,36 @@
+CREATE TABLE `ban` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `player_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `author_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `reason` VARCHAR(255) NOT NULL,
+  `expires_at` TIMESTAMP NULL DEFAULT NULL COMMENT 'if null the ban is permanent',
+  `level` ENUM('CHAT', 'GLOBAL') NOT NULL,
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
+  FOREIGN KEY (`player_id`) REFERENCES `login`(`id`),
+  FOREIGN KEY (`author_id`) REFERENCES `login`(`id`)
+) COMMENT = 'ban can be revoked, see ban_revoke';
+
+CREATE TABLE `ban_revoke` (
+  `ban_id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `reason` VARCHAR(255) NOT NULL,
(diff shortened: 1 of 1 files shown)


Brutus5000:
The `ban_disable_data` is somehow disposable.
It's supposed to be a 1:1 relationship.
Why do we need a new primary key?
If revocation of a ban is the only "change" possible, you need neither the update_time in ban, nor in ban_disable_data.

I suggest to add a revoked_by, revoke_reason, revoke_time columns into the ban table instead.
Don't split information that belongs together.

IDragonfire:
My first suggestion was with additional fields... But then we have a huge amount of of null values...

Brutus5000:
This is common sense in databases and has no impact on performance or disk usage. So no valid argument for me.

IDragonfire:
I will rename it to `revoke` ;) Better name^^

@Brutus5000 Have you some sources for the performance stuff? I remember the opposite ... many `null` values are bad:
http://stackoverflow.com/a/12321323

But you are right ... I should use a real 1:1 releation:
https://www.mkyong.com/mysql/how-to-define-one-to-one-relationship-in-mysql/


 `update_time` can also change if you update the reason ... 
So if I merge the `revoke` I need
* `revoke_author`
* `revoke_reson`
* `revoke_create_time`
* `revoke_update_time`

@micheljung What is your suggestion?
PS.: I use also a wrong 1:1 relation for the clan structure: 
https://github.com/FAForever/db/blob/develop/migrations/V15__fix-clan-structure.sql#L23

Brutus5000:
This is my experience working with production systems.
Hell, SAP even uses strings as Primary Key for everything and it's no performance issue for systems with millions of entries. I don't think we will run into any problems.

Of course 1:1 table relationships are allowed, but should only be used in a manner where it actually has many of entries in on table, but few in the other. Compared to other tables we'll have few entries in both.

Sheeo:
I don't recommend allowing `expires_at` to be `null`, because having exactly this previously caused all permanent bans to cease functioning when a contributor didn't know that was what that value meant.

The comment wouldn't likely have fixed that. Just having "permanent" mean 999 years isn't a problem?

Sheeo:
Instead of having a join table for revocation, since we're already reworking bans -- why not just make the ban table have a field `ban_level`, with an enum (Optionally a bit-vector) value like:

```
NOT_BANNED // Used for revocation
CHAT_BANNED // New status just for those who are chat-banned, could be used to disable ingame-chatting or more sophisticated banning behavior
GAME_HOST_BANNED // Cannot host games
GAME_BANNED // Cannot play games at all
BANNED // All of the above
```

IDragonfire:
This is against 3NF because not banned status  depend on expire on and current time ;) 

The open question is only how to resolve the 1:1 relations... 

Merge tables or two separate tables....

IDragonfire:
I wrote already the api... Status with not banned is an transient property, calculated on the fly... 
I can also write a view with all active bans...

Sheeo:
Nobody cares whether your database conforms to 3NF or not.

I don't particularly care about the 1:1 correlation, but I'm not against denormalising tables for simplicity. Don't be an extreme 3NF'er just for the sake of it, and don't keep obtuse denormalised tables. Best of both worlds.

IDragonfire:
Any updates?

IDragonfire:
@micheljung Anything else until this can be merged?

micheljung:
I don't have strong feelings about this. Looks good to me. If nobody opposes, I'll merge it soon.

PR #86, opened 2017-04-25
Fix #85 timestamp columns cannot be null

### LINKED ISSUE #85: MySQL Bug Cannot Insert null timestamps
Expected behaviour
https://bugs.mysql.com/bug.php?id=72628

```sql
mysql> create table t1(c1 timestamp not null default '2000-01-01 00:00:00');
Query OK, 0 rows affected (0.02 sec)

mysql> insert into t1 values();
Query OK, 1 row affected (0.00 sec)

mysql> insert into t1 values(null);
Query OK, 1 row affected (0.01 sec)

mysql> insert into t1 values(default);
Query OK, 1 row affected (0.00 sec)

mysql> select * from t1;
+---------------------+
| c1                  |
+---------------------+
| 2000-01-01 00:00:00 |
| 2014-08-01 06:11:50 |
| 2000-01-01 00:00:00 |
+---------------------+
3 rows in set (0.00 sec)
```

Behaviour with `5.7.16`
```sql

mysql> create table t1(c1 timestamp not null default '2000-01-01 00:00:00');
Query OK, 0 rows affected (0.22 sec)

mysql> insert into t1 values(null);
ERROR 1048 (23000): Column 'c1' cannot be null
```

IDragonfire:
Version upgrade from `5.7.16` to `5.7.17` or `5.7.18` solved the problem ...

no real idea why
https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-17.html

Fix #85 timestamp columns cannot be null

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM mysql:5.7
+FROM mysql:5.7.18
 
 ENV FLYWAY_VERSION 4.0.3
 


IDragonfire:
@duk3luk3 @micheljung This PR fixed the version of the daatabase that every developer has the same version ...
Currently this Pr is not needed, because if you create a new docker version , mysql version `5.7.17` is used (now idea why `5.7.18` is not used) ...

duk3luk3:
I think pinning the version is good in this case and the PR should be merged.

PR #90, opened 2017-04-28
[8] ACU Skin System

IDragonfire:
Open for discussion

- [x] How to select the current skin? not in db, lobby onle
- [x] How to link sim model to db model (column `id` enough?)
- [x] How to prevent users to change default skin (`checksum` enough?) -> checkum not neeed

More details: https://youtu.be/fCU5kL8WDlM

### Please consider the affect on the following components
This PR only added new stuff:
- [ ] python-server: need implementation
- [x] python-api: not planned
- [ ] java-api: need implementation
- [x] website: not planned

### DIFF
--- a/migrations/V37__acu-skins.sql
+++ b/migrations/V37__acu-skins.sql
@@ -1,0 +1,22 @@
+CREATE TABLE `acu_skin` (
+  `id` mediumint(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `faction` ENUM('UEF', 'CYBRAN', 'AEON', 'SERAPHIM') NOT NULL,
+  `name` varchar(40) CHARACTER SET 'utf8' NOT NULL COMMENT 'display name, e.g. for modapp',
+  `comment` text CHARACTER SET 'utf8' COMMENT 'optional comment about the skin, e.g. purpose',
+  `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `acu_skin_assignment` (
+  `id` mediumint(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `skin_id` mediumint(8) UNSIGNED NOT NULL,
+  `player_id` mediumint(8) UNSIGNED NOT NULL,
+  `comment` text CHARACTER SET 'utf8' COMMENT 'optional comment about the skin assignment, e.g. why the user received the skin',
+  `expires_at` DATETIME NULL DEFAULT NULL COMMENT 'If null, the avatar is permanent',
+  `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
+  FOREIGN KEY (`skin_id`) REFERENCES `acu_skin`(`id`),
+  FOREIGN KEY (`player_id`) REFERENCES `login`(`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
--- a/test-data.sql
+++ b/test-data.sql
@@ -140,3 +140,13 @@ insert into updates_faf_files (id, fileId, version, name, md5, obselete) values
     (680, 12, 3656, 'env_0.3656.nxt', '32a50729cb5155ec679771f38a151d29', 0);
 
 insert into teamkills (teamkiller, victim, game_id, gametime) VALUE (1, 2, 1, 3600);
+
+-- sample acu skins
+insert into acu_skin(id, faction, name, comment) values
+    (2, 'CYBRAN', 'Red Cybran Skin', 'We all love Cybran'),
+    (4, 'UEF', 'Blue Humans', null),
+    (9, 'AEON', 'Green Aliens', 'For building Paragons');
+
+insert into acu_skin_assignment(skin_id, player_id, expires_at, comment) values
+  (2, 2, null, 'This skin will you get forever'),
+  (2, 4, NOW(), 'You should no longer use or select this skin, it is expired');


IDragonfire:
@JJ173 suggested `skin_1_uef` as `sim_name` to identify the skin ...

JJ173:
Client needs to read all the skins from the DB by ID or name. I'd like to suggest that however they are read, the name should be obtained. 

E.G ID 1 = 1_uef_test

The client should send the skin name to the game (skins are given by moderators) for each player so they can select it using a similar system to the mod manager. I can handle the sim code from there. If I get a test skin ready, I'd like to get a basic system running. Can this be approved? Skins can't be modified as they'll be stored on a server.

JJ173:
I'd like to add a skin to test this prototype. I'll get one ready and post it here, and then I'll start work on the client stuff.

duk3luk3:
Some possible extensions/comments:

* Should skins be able to have an expiry date?
* Should skin assignments have a comment (e.g. "for winning tourney on xx.xx.xx")
* A player can have more than one skin, right? How to determine which one is active?
* Why is the checksum needed?

JJ173:
I don't think Skins should have expiry dates. Similar to avatars, they will be rewarded for either donation, contribution or winning a tournament. I believe a skin should have a comment to allow a user to view what they were rewarded for. A player can be given more than one skin, the active skin will be handled in the game lobby. We need some sort of method to prevent users from cheating the system. I assume that's why we need a checksum.

duk3luk3:
> Similar to avatars, they will be rewarded for either donation, contribution or winning a tournament

Tournament avatars are supposed to expire.

> We need some sort of method to prevent users from cheating the system

Well, presumably the player selects their skin in the lobby. This results in a GPGNet message sent to the server. The server can then check that the selection is valid and send a GPGNet message back to all clients.

Alternatively, clients can check by themselves using the API as well.

I suppose players could trick people into using sim mods that change the skin. I think those sim mods could also circumvent any checksum check.

I think this needs to be checked by whoever will be implementing the skins into fafmod.

JJ173:
Then we will set an expiry on Tournament skins. Other than that, I don't see many issues. I'm unaware of the methods to check for changes via GPGNet.

IDragonfire:
checksum was a simple solution to prevent simple cheating, e.g. just replace the default uncool skin with an cool skin ...

aeoncleanse:
GPGnetsend messages are probably the best bet. Not really my area though.

IDragonfire:
skin selection is a little bit complicated, because you can select a skin for every faction  ...
# Suggestion
A `order` column ... the first matching skin with the highest order is used ...

JJ173:
If skins are stored in `textures.nx2` wouldn't they just be re-downloaded if someone was to make a local modification?

duk3luk3:
As was discussed in council/dev conference last night:

* checksum is unneeded
* skin expiry is needed

Please make a decision about 

* Should skin assignments have a comment (e.g. "for winning tourney on xx.xx.xx")

And then update this PR so it can be put on the test server, and so that implementation work in the API can begin.

IDragonfire:
> And then update this PR so it can be put on the test server, and so that implementation work in the API can begin.

Coding on the API can begin without pulling this PR^^

JJ173:
A user will be able to select a skin in the lobby. Being able to select a skin for each faction isn't a bad thing, as they only play one faction. I'll work on how to get that done, it's not relevant to the PR.

Skins should have a caption to say what they have been given for like the example @duk3luk3 provided. That should be stored with the skin in the DB.

JJ173:
Maybe a `faction` column would be handy to check what skin belongs to which faction (if possible)

duk3luk3:
I think we need a filename column.

IDragonfire:
@duk3luk3 I have no idea about this , but the filename can be the id of the skin in the database.... It was my first question and @JJ173 accepted it .... 

Ps.: It is a bad idea to encode informations in the filename, so for me an id is perfect. Downside is, that it is not as readable ...

JJ173:
ID would probably be better, as long as the name of the skin can be provided somewhere. Maybe like a title or filename as you say. So the test skin ID can be `1`, the filename `UEF_Test_Skin` and the Title `UEF Tester Skin`. Some spiffy title that I can include in the manager.

migrations/V37__acu-skins.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE `acu_skin` (
+  `id` mediumint(8) UNSIGNED NOT NULL AUTO_INCREMENT,
```
ChrisKitching: Limiting yourself to 256 skins maximum seems like something you may regret. It doesn't hurt to spend a few extra bits to make sure you don't have a nightmare headache later :D

IDragonfire: `16777215` values are enough https://dev.mysql.com/doc/refman/8.0/en/integer-types.html
the number behind is for the minimal display width in mysql:
https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html

axel1200:
What is the status of this here ???

1-alex98:
seems dead

migrations/V37__acu-skins.sql
norraxx: @ChrisKitching this 8 near mediumint is used for printing. Imho zerofill is using that value, otherwise it's value is mediumint

migrations/V37__acu-skins.sql
```diff
@@ -0,0 +1,22 @@
...
+  `skin_id` mediumint(8) UNSIGNED NOT NULL,
+  `player_id` mediumint(8) UNSIGNED NOT NULL,
+  `comment` text CHARACTER SET 'utf8' COMMENT 'optional comment about the skin assignment, e.g. why the user received the skin',
+  `expires_at` DATETIME NULL DEFAULT NULL COMMENT 'If null, the avatar is permanent',
+  `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
+  FOREIGN KEY (`skin_id`) REFERENCES `acu_skin`(`id`),
+  FOREIGN KEY (`player_id`) REFERENCES `login`(`id`)
```
norraxx, line 21: From the usage that i see, it's better to add coverage indexes, like: 
```SQL
  FOREIGN KEY (`player_id`) REFERENCES `login`(`id`),
  INDEX (`login_id`, `expires_at`, `skin_id`)
```
Reason is simple, you may or will join this table on login, on non expired, and what you need is skin_id as result. No need to open whole table, because if each of faf player will get one skin per fraction, it'll be 4x 800000.

PR #92, opened 2017-04-28
Feature/fix migration 26

IDragonfire:


Brutus5000:
On a side note: The naming of your branch makes no sense. It's either fix or feature, but probably not both.

micheljung:
Except that only `feature/` is a valid prefix and `fix/` *is not*. `bugfix/` would be.

IDragonfire:
its a cool feature to have a fix ^^ but you are right xD

PR #97, opened 2017-04-28
Fix #96 DB Docker not working on window

### LINKED ISSUE #96: DB Docker not working on window
image

IDragonfire:
Fix #96 DB Docker not working on window


duk3luk3:
Can you please explain what the problem is and how this PR fixes it?

IDragonfire:
If you open the issue you should get a good picture ...
On Window you have a different line ending and docker (linux) don't like this ...

duk3luk3:
That sounds like it should be fixed by setting up your git config to use linux newlines, or rather, not change newlines at all.

IDragonfire:
A repository should work ...
`.gitattribute` is for such use cases ...

I want not to edit my global or project specific git config for every computer and repository xD

We should make it as easy as possible  ...
Otherwise you must write ...

# Window Installation Guide
1. Before cloning please add this to your git config

PR #98, opened 2017-05-02
Remove ON UPDATE CURRENT_TIMESTAMP from game_stats

### LINKED ISSUE #83: Remove CURRENT_TIMESTAMP for game_stats.startTime
`game_stats.startTime` is currently set implicitly by `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` which, if you think about it for one second, is not a smart thing to do.

This causes any game that have any of its values updated (like `validity`) when the game ends to have its `startTime` set to the game end time. Outch!

duk3luk3:
game_stats.startTime had a nonsensival ON UPDATE CURRENT_TIMESTAMP
condition.

Fixes #83.


migrations/V30__game_stats-starttime-timestamp.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE faf.game_stats CHANGE startTime startTime TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```
micheljung: The schema `faf` should not be part of the script, and the script should be named `V30__game-stats-starttime-timestamp.sql` (dash instead of underscore)

duk3luk3: Why not? Also, it's part of the V29 migration.

micheljung: Because then it doesn't work if the DB is named differently (like `faf_lobby`, on the prod server).
It shouldn't be in V29 either, I'm glad you saw that!

PR #100, opened 2017-05-02
Fix #99 make game_stats.score nullable

### LINKED ISSUE #99: Allow 'game_player_stats.score' to be null
`score` is currently non-null and initialized with `-1` by the server which basically means "no score".
Instead, `score` should be nullable. Since the server sets it explicitly anyway and is the only software that sets it, we don't need to fix anything on the server.

micheljung:


duk3luk3:
They will be migrated once the server is updated to support it, of course.

PR #101, opened 2017-05-02
Fix #87 add featured mod constraint

### LINKED ISSUE #87: Foreign key for game_stats.gameMod -> game_featuredMod.id
We have games that have been played with mods that are no longer in our database.
This should never ever happen again.

micheljung:
Removes all games for which we lost the "featured mod" information
(some featured mods were removed with the March 2017 server update) and
and adds a constraint so this may never happen again.

### DIFF
--- a/migrations/V32__featured-mod-constraint.sql
+++ b/migrations/V32__featured-mod-constraint.sql
@@ -1,0 +1,16 @@
+# Estimated script run time: 25min
+
+# As voted for by Sheeo, Brutus5000, and Yenon, we delete games we no longer have featured mods for.
+# Alternatively, we'd have to create dummy featuredMod entries.
+DELETE FROM game_player_stats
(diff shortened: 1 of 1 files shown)


IDragonfire:
How you calculate `Estimated script run time: 25min`?

duk3luk3:
What's the impact of throwing away those games? How recent are they? I assume they can't be replayed anymore? March 2017 is pretty recent, don't we have db dumps from before the update?

While we're at it we should also think about a way to "remove" featured mods that doesn't cause issues like this in the first place - i.e. hiding them using a flag instead of removing them.

micheljung:
Sorry I meant march 2016. There's already a way to hide them, no idea why they were deleted. The impact is that about 250k games and their player stats will be deleted. No impact on rating or games played.
The time estimation comes from a local test run.

PR #117, opened 2017-05-15
[0] Implement `clan_tag` view for python server

IDragonfire:
Implement `clan_tag` view for python server

https://github.com/FAForever/server/blob/a5adc6f441ea49b3e5167eed0e1414bb322af3fa/server/player_service.py#L70


IDragonfire:
Drop views and use direct queries on the server to have more control over the query,
potential performacne issue if the view use ` temptable algorithm`:
https://dev.mysql.com/doc/refman/5.7/en/view-restrictions.html

PR #127, opened 2017-07-24
Fix test-data.sql (constraint issues)

Brutus5000:
Begin import of test-data with cleaning constraint-affected tables.

### DIFF
--- a/test-data.sql
+++ b/test-data.sql
@@ -1,7 +1,37 @@
 -- DUMMY DATA ONLY, FOR USE IN UNIT TESTS
 
+
+-- regular deletes without constraints
+DELETE FROM uniqueid;
+DELETE FROM unique_id_users;
+DELETE FROM global_rating;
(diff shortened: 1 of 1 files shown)


IDragonfire:
Now we delete twice ... not a good style ...

test-data.sql
```diff
@@ -1,56 +1,73 @@
...
+DELETE FROM avatars;
+DELETE FROM avatars_list;
+DELETE FROM ban_revoke;
+DELETE FROM ban;
+DELETE FROM clan_membership;
+DELETE FROM clan;
+DELETE FROM game_player_stats;
+DELETE FROM game_stats;
+DELETE FROM teamkills;
+DELETE FROM login;
```
IDragonfire: `login` should be deleted at last ;)

Brutus5000:
I have reversed the order. Now the constraint-restricted tables are deleted last (with login being the very last).

PR #138, opened 2017-10-20
Create /root/.my.cnf with db credentials

duk3luk3:
Allows exec'ing into mysql without having to pass credentials

Having database passwords on the commandline and even in scripts is bad. This way the password is read from the config file.

This doesn't make anything any less secure, since the db root pw is in an env var inside the container anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,4 +39,11 @@ RUN chmod +x dump-structure.sh
 # Copy test data for CI
 COPY test-data.sql /
 
+# Copy pre-init script
+COPY pre-entry.sh /usr/local/bin/docker-pre-entry.sh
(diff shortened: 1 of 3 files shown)


duk3luk3:
@micheljung @Brutus5000 any idea what caused that test failure?

Brutus5000:
I'm not sure how your entrypoint works, but if it works I would assume lines like this from `test-migrations.sh`:
`docker exec -i faf-db mysql -uroot -pbanana <<< "DROP DATABASE faf;"`

need to be changed?

duk3luk3:
What you wrote was very confusing but gave me the right inspiration :P

PR #147, opened 2017-12-05
Moderator notes per user

Brutus5000:
As requested in https://github.com/FAForever/faf-moderator-client/issues/16


migrations/V48__notes-per-user.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE user_notes
+(
+    id INT(11) PRIMARY KEY NOT NULL AUTO_INCREMENT,
+    user_id MEDIUMINT(9) unsigned NOT NULL,
+    author MEDIUMINT(9) unsigned NOT NULL,
+    watched TINYINT(4) DEFAULT '0' NOT NULL,
```
micheljung: I'm not sure what `watched` means, exactly. A comment would help.

PR #151, opened 2018-04-11
Voting

### LINKED ISSUE #150: implement voting

axel1200:
Fixes #150

### DIFF
--- a/migrations/V51__voting.sql
+++ b/migrations/V51__voting.sql
@@ -1,0 +1,80 @@
+CREATE TABLE IF NOT EXISTS voting_subject (
+  `id`                 INT UNSIGNED     NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `subject_key`        VARCHAR(255)     NOT NULL
+  COMMENT 'With this key the ''subject'' can be loaded from the messages table',
+  `begin_of_vote_time` DATETIME         NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `end_of_vote_time`   DATETIME         NOT NULL,
+  `min_games_to_vote`  INT UNSIGNED     NOT NULL DEFAULT 0
+  COMMENT 'The number of games a player must have to in order to vote',
+  `description_key`    VARCHAR(255)
+  COMMENT 'The ''description-key'' of the voting subject with it additional information (HTML) can be loaded from the messages table',
+  `topic_url`          TEXT
+  COMMENT 'An URL to a page with additional information, e.g. a forum post.',
+  `reveal_winner`      TINYINT UNSIGNED NOT NULL DEFAULT 0
+  COMMENT 'When set to true after voting ended the winner is calculated and shown publicly',
+  `create_time`        TIMESTAMP        NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`        TIMESTAMP        NOT NULL DEFAULT CURRENT_TIMESTAMP
+  ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS voting_question (
+  `id`                 INT UNSIGNED     NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `max_answers`        INT UNSIGNED     NOT NULL DEFAULT 1,
+  `question_key`       VARCHAR(255)     NOT NULL
+  COMMENT 'With this key the ''question'' can be loaded from the messages table',
+  `voting_subject_id`  INT UNSIGNED     NOT NULL,
+  CONSTRAINT voting_subject_of_voting_question FOREIGN KEY (`voting_subject_id`) REFERENCES voting_subject (id),
+  `description_key`    VARCHAR(255)
+  COMMENT 'The ''description-key'' of the voting question, with it additional information (HTML) can be loaded from the messages table',
+  `ordinal`            INT UNSIGNED     NOT NULL DEFAULT 0
+  COMMENT 'The ''ordinal'' of the voting question, in the client questions are shown according to their ordinal value',
+  `alternative_voting` TINYINT UNSIGNED NOT NULL DEFAULT 0
+  COMMENT 'Defines whether alternative voting applies to this question',
+  `create_time`        TIMESTAMP        NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`        TIMESTAMP        NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS voting_choice (
+  `id`                 INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `choice_text_key`    VARCHAR(255) NOT NULL
+  COMMENT 'With this key the ''choice_text'' can be loaded from the messages table',
+  `voting_question_id` INT UNSIGNED NOT NULL,
+  CONSTRAINT voting_question_of_voting_choice FOREIGN KEY (`voting_question_id`) REFERENCES voting_question (id),
+  `description_key`    VARCHAR(255) NOT NULL
+  COMMENT 'The ''description-key'' of the voting choice, with it additional information (HTML) can be loaded from the messages table',
+  `ordinal`            INT UNSIGNED NOT NULL DEFAULT 0
+  COMMENT 'The ''ordinal'' of the voting choice, in the client choices are shown according to their ordinal value',
+  `create_time`        TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`        TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS vote (
+  `id`                INT UNSIGNED       NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `player_id`         MEDIUMINT UNSIGNED NOT NULL,
+  CONSTRAINT player_of_vote FOREIGN KEY (`player_id`) REFERENCES login (id),
+  `voting_subject_id` INT UNSIGNED       NOT NULL,
+  CONSTRAINT voting_subject_of_vote FOREIGN KEY (`voting_subject_id`) REFERENCES voting_subject (id),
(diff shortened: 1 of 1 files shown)


migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
...
+  `question`          TEXT         NOT NULL,
+  `voting_subject_id` INT UNSIGNED NOT NULL,
+  CONSTRAINT voting_subject_of_voting_question FOREIGN KEY (`voting_subject_id`) REFERENCES faf.voting_subject (id),
+  `description`       TEXT
+  COMMENT 'The ''description'' of the voting question, here additional information can be stored',
+  `create_time`       TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`       TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE faf.voting_choice (
```
Brutus5000: I strongly suggest an additional sorting order column for the choices.

migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
...
+  `id`                 INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `name`               TEXT         NOT NULL,
```
Brutus5000: It's probably the actual choice question, also rather text than a name? Description is additional.

migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
+CREATE TABLE faf.voting_subject (
+  `id`                 INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `name`               VARCHAR(255) NOT NULL,
+  `begin_of_vote_time` TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `end_of_vote_time`   TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
micheljung: If someone creates a voting subject without an end time, the end time will be set to "now" - giving it an actual lifetime of 0 seconds.

I think it should be non-null and have no default.

migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
...
+  `min_games_to_vote`  INT UNSIGNED NOT NULL DEFAULT 0
+  COMMENT 'The number of games a player must have to in order to vote',
+  `description`        TEXT
+  COMMENT 'The ''description'' of the voting subject',
```
micheljung: Oh really, `voting_subject.description` is the `description` of the voting subject? You don't say 😄 

If anything, I'd like to know whether this should be plain text, HTML, Markup or something else. However, this is none of the database's business because it's an implementation detail of the application. Which makes me think whether we should add a `description_type` which can be set to a MIME-Type like `text/plain`, `text/html` or `text/markup` but maybe I'm overengineering things here.

Then again, we have exactly this problem with featured mods; some of their descriptions are in HTML while others were written as plain text - resulting in either of them being rendered incorrectly in the client.

Therefore, I would add `description_type` after all.

migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
...
+  `topic_url`          TEXT
+  COMMENT 'A link going for example to the forum or another website, with the related threador additional information',
```
micheljung: De-facto limit of URLs is 2000 characters, so I'd make this a `varchar(2000)` rather than `text`.

Description: `An URL to a page with additional information, e.g. a forum post.`

and since not every subject necessarily needs an URL, `NOT NULL`.

migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
+CREATE TABLE faf.voting_subject (
```
micheljung: The script shouldn't assume a database name. In fact, the database is called `faf_lobby` in production, so remove all occurrences of `faf.`

migrations/V50_voting.sql
```diff
@@ -0,0 +1,61 @@
...
+);
+
+CREATE TABLE faf.voting_question (
+  `id`                INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `max_answers`       INT UNSIGNED NOT NULL DEFAULT 1,
+  `question`          TEXT         NOT NULL,
+  `voting_subject_id` INT UNSIGNED NOT NULL,
+  CONSTRAINT voting_subject_of_voting_question FOREIGN KEY (`voting_subject_id`) REFERENCES faf.voting_subject (id),
+  `description`       TEXT
+  COMMENT 'The ''description'' of the voting question, here additional information can be stored',
```
micheljung: See my comment on subject description

migrations/V50_voting.sql
axel1200: description_type is supposed to be a column?

axel1200: Yeah maybe you are overengennering, because that would make the client implemtation a lot more complicated. Imagine, we would then also be so generic that we need to show webview/text whatever/ mark up webview depending on the type and then in the end people set the wrong type. I think we should just make it html or plain text and decide on it.

axel1200: I would chose html for most descriptions, everybody can make it html if he uses a online html editor https://html-online.com/editor/

migrations/V50_voting.sql
```diff
@@ -0,0 +1,65 @@
+CREATE TABLE voting_subject (
+  `id`                 INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `subject_text`       VARCHAR(255) NOT NULL,
```
micheljung: what's the difference between a subject and a subject text?

micheljung: I'd just call it `subject`

migrations/V50_voting.sql
```diff
@@ -0,0 +1,65 @@
...
+  `ordinal`            INT UNSIGNED NOT NULL DEFAULT 0
+  COMMENT 'The ''ordinal'' of the voting choice, in the client choices are shown according to their ordinal value',
+  `create_time`        TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`        TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE vote (
+  `id`                INT UNSIGNED       NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `create_time`       TIMESTAMP          NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`       TIMESTAMP          NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
micheljung: I think we should always put these at the end, like in all other tables

migrations/V50_voting.sql
```diff
@@ -0,0 +1,65 @@
...
+  CONSTRAINT player_of_vote FOREIGN KEY (`player_id`) REFERENCES login (id),
+  `voting_subject_id` INT UNSIGNED       NOT NULL,
+  CONSTRAINT voting_subject_of_vote FOREIGN KEY (`voting_subject_id`) REFERENCES voting_subject (id),
+  UNIQUE KEY `one_vote_only` (`player_id`, `voting_subject_id`)
+);
+
+CREATE TABLE voting_answer (
+  `id`                 INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `create_time`        TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`        TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
micheljung: same

migrations/V50_voting.sql
```diff
@@ -0,0 +1,65 @@
...
+  `vote_id`            INT UNSIGNED NOT NULL,
+  CONSTRAINT vote_of_voting_answer FOREIGN KEY (`vote_id`) REFERENCES vote (id),
+  `voting_question_id` INT UNSIGNED NOT NULL,
+  CONSTRAINT voting_question_of_voting_answer FOREIGN KEY (`voting_question_id`) REFERENCES voting_question (id),
+  `voting_choice_id`   INT UNSIGNED NOT NULL,
+  CONSTRAINT voting_choice_of_voting_answer FOREIGN KEY (`voting_choice_id`) REFERENCES voting_choice (id),
+  UNIQUE KEY `one_answer_only` (`vote_id`, `voting_choice_id`)
+);
```
micheljung: This symbol means there's a missing newline at the end, which Git prefers. You can have them added automatically in the commit dialog i think

axel1200:
Dont we want the key thingy here as well?

axel1200:
I am going for tutorials first

migrations/V50__voting.sql
```diff
@@ -0,0 +1,65 @@
+CREATE TABLE voting_subject (
+  `id`                 INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `subject`            VARCHAR(255) NOT NULL,
+  `begin_of_vote_time` TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `end_of_vote_time`   TIMESTAMP    NOT NULL,
+  `min_games_to_vote`  INT UNSIGNED NOT NULL DEFAULT 0
+  COMMENT 'The number of games a player must have to in order to vote',
+  `description`        TEXT
+  COMMENT 'The ''description'' of the voting subject, specifying further information in HTML',
+  `topic_url`          VARCHAR(2000)
```
ChrisKitching: 2000 characters for a URL?!

If you really need 2000 characters, this should be of type TEXT. very large VARCHARs aren't that efficient.

germanicianus:
’’’
The HTTP protocol does not place any a priori limit on the length of a URI. Servers MUST be able to handle the URI of any resource they serve, and SHOULD be able to handle URIs of unbounded length if they provide GET-based forms that could generate such URIs. A server SHOULD return 414 (Request-URI Too Long) status if a URI is longer than the server can handle (see section 10.4.15).

      Note: Servers ought to be cautious about depending on URI lengths
      above 255 bytes, because some older client or proxy
      implementations might not properly support these lengths.
’’’
https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.2.1

PR #153, opened 2018-05-10
Tutorials featured mod

### LINKED ISSUE #148: Tutorial maps
Add tables for tutorial maps

axel1200:
Fixes #148

### DIFF
--- a/migrations/V50__tutorial-featured-mod.sql
+++ b/migrations/V50__tutorial-featured-mod.sql
@@ -1,0 +1,4 @@
+CREATE TABLE updates_tutorials_files
+  LIKE updates_faf_files;
+CREATE TABLE updates_tutorials
+  LIKE updates_faf;


migrations/V50__tutorial-featured-mod.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE `updates_tutorials_files` (
+  `id`       INT(10) UNSIGNED     NOT NULL AUTO_INCREMENT,
+  `fileId`   SMALLINT(5) UNSIGNED NOT NULL,
+  `version`  INT(11)              NOT NULL,
+  `name`     VARCHAR(45)          NOT NULL,
+  `md5`      VARCHAR(45)          NOT NULL,
```
ChrisKitching: md5 checksums are 32 characters long, so this doesn't need to be 45 characters.

Also note that `CHAR()` fields are cheaper than `VARCHAR()` in cases where the exact length is known. `VARCHAR(X)` fields are for when you want a string of up to `X` characters (and it'll dynamically only allocate that much, plus some bookkeeping). For this field you know it will always be exactly 32, so you should use `CHAR(32)`, allowing the storage engine to statically allocate 32 chars and not bother with any of the bookkeeping.

migrations/V50__tutorial-featured-mod.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE `updates_tutorials_files` (
+  `id`       INT(10) UNSIGNED     NOT NULL AUTO_INCREMENT,
+  `fileId`   SMALLINT(5) UNSIGNED NOT NULL,
```
ChrisKitching: A maximum value of 32 seems... risky. A couple more bits won't cost anything significant and may well save you from a future schema change.

PR #155, opened 2018-06-22
Implement user groups and permissions

Brutus5000:
This is my new approach on the permission system, based on Dragonfires initial PR.

The original role is reinterpreted as a hierachical structure to reflect the FAF organization.
Gone would be the times when people needed to paint organizational charts and post them in the forums, which are out-of-date in the blink of an eye.

Let's make our organization queryable via API and (which would be neat) embedded in the website and the client.
This would be my initial setup:

```
| FAF [root group]
|-- Administration councilor
|-- M&M counselor
|--|-- Map vault team
|--|-- Mod vault team
|--|-- Mappers
|--|-- Modders
|-- Player councilor
|--|-- Balance team
|--|-- Tournament directors
|--|-- Tutorials team
|-- Moderation councilor
|--|-- Global moderators
|--|-- IRC moderators
|--|-- Forum moderators
|--|-- Discord moderators
|-- Promotion councilor
|--|-- News posters
|--|-- News translators
|--|-- Casters
|--|-- Streamers
|-- Software councilor
|--|-- Server administration
|--|-- Python client developers
|--|-- Java client developers
|--|-- GW developer developers
|--|-- Web developers
|--|-- ...
```

Now permissions can be assigned to any of the groups (but they are not supposed to be inheritable) and directly assigned members of the group inherit it.
For special permissions outside of the official organizations groups can be "not public", then they are not queryable.

If we can agree on this, I will implement it in the java-api and built a configuration ui in the moderator client.
Users in a group should be able to manage permissions of their subgroups. In the proposed structure every counselor could hand out permissions for his sub-teams.

I suggest a slow transition - remove the old lobby_groups only if everything is ported.


migrations/V52__permission-system.sql
```diff
@@ -0,0 +1,42 @@
+CREATE TABLE `user_group` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `name` VARCHAR(40) NOT NULL COMMENT 'e.g. Admin, Mod, Council of Setons',
+  `parent_group_id` MEDIUMINT(8) UNSIGNED,
+  `public` TINYINT(1)  NOT NULL COMMENT 'Public groups are visible for everyone, the rest only for internal permissions',
```
norraxx: Could data be filtered by public/parent_group_id? If yes, how about the index + foreign key on parent_group_id?

migrations/V52__permission-system.sql
```diff
@@ -0,0 +1,42 @@
...
+  `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`) )
+COMMENT = 'List of all user groups. Some of them are only informative  (i.e. council of setons on website), others for permissions system.';
```
Sheeo: You'll want a local foreign key on `parent_group_id`:

```sql
FOREIGN KEY `parent_group_key` (`parent_group_id`)
REFERENCES `user_group` (`id`)
ON DELETE CASCADE
ON UPDATE SET NULL;
```

This way `parent_group_id` is always valid if set.

PR #156, opened 2018-07-01
#194 faf-java-api speed up achievement_statistics view

norraxx:
Achievments take long to load

### DIFF
--- a/migrations/V52__view-remake.sql
+++ b/migrations/V52__view-remake.sql
@@ -1,0 +1,79 @@
+-- small indexes for achievement_definitions and login, no need to load more
+ALTER TABLE `achievement_definitions`
+ADD UNIQUE `id_create_time` (`id`, `create_time`);
+ALTER TABLE `login`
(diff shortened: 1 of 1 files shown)


Brutus5000:
-- DO NOT MERGE --
Slightly modified test run on production shows that it is slower than the old view.
Needs further investigation.

norraxx:
I'm ok with that.

How about to leave just one index, that might relieve DB during query?
I'll check if it does so, ofcourse.
```
ALTER TABLE `player_achievements`
ADD INDEX `state_player_id_achievement_id_update_time` (`state`, `achievement_id`,  `player_id`, `update_time`);
```

Brutus5000:
How do we want to proceed with this PR? As our tests on the test server indicate the new indices itself speed up the query a bit, but the new query itself is not fast than the old one.

axel1200:
We fixed it right?

micheljung:
Closing due to inactivity

PR #157, opened 2018-07-03
Added a table for moderation reports

Rackover:
Added a table which will store moderation reports, completed or not.

### DIFF
--- a/migrations/V52__moderation_report.sql
+++ b/migrations/V52__moderation_report.sql
@@ -1,0 +1,46 @@
+CREATE TABLE IF NOT EXISTS moderation_report (
+  `id`                      INT UNSIGNED            NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `reporter_id`             MEDIUMINT UNSIGNED      NOT NULL,
+  `report_description`      TEXT,
+ 
+  `game_id`                 INT UNSIGNED            DEFAULT NULL
+  COMMENT 'If equal to NULL, the offense did not happen ingame',
+
+  `report_status`           ENUM('AWAITING', 'PROCESSING', 'COMPLETED', 'DISCARDED')   NOT NULL
+  COMMENT 'The current status of the report. This will be accessable by the concerned user.',
+  
+  `game_incident_timecode`  VARCHAR(100)
+  COMMENT 'The timecode of the incident ingame, indicated by the user in their own terms',
+  
+  `moderator_notice`        TEXT
+  COMMENT 'A public notice left by the moderator which will be addressed to the reporter once the report is marked as either COMPLETED or DISCARDED',
+  `moderator_private_note`  TEXT
+  COMMENT 'A private notice left by the moderator which any other moderator accessing the record will be able to see.',
+  `last_moderator`          MEDIUMINT UNSIGNED      DEFAULT NULL
+  COMMENT 'Last moderator to have updated the report',
+  
+  `create_time`             TIMESTAMP              NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`             TIMESTAMP              NOT NULL DEFAULT CURRENT_TIMESTAMP  ON UPDATE CURRENT_TIMESTAMP,
+  
+  CONSTRAINT game_id_of_moderation_report FOREIGN KEY (`game_id`) REFERENCES game_stats (id),
+  CONSTRAINT reporter_id_of_moderation_report FOREIGN KEY (`reporter_id`) REFERENCES login (id),
+  FOREIGN KEY (`last_moderator`) REFERENCES login (id)
+);
+
+CREATE TABLE IF NOT EXISTS reported_user (
+  `id`                     INT UNSIGNED             NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `player_id`              MEDIUMINT UNSIGNED        NOT NULL,
+  `report_id`              INT UNSIGNED             NOT NULL,
+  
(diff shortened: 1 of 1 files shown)


migrations/V52__moderation_report.sql
```diff
@@ -0,0 +1,40 @@
...
+  `last_moderator`        MEDIUMINT UNSIGNED     DEFAULT NULL
+  COMMENT 'Last moderator to have updated the report',
+  
+  CONSTRAINT game_id_of_moderation_report FOREIGN KEY (`game_id`) REFERENCES game_stats (id),
+  CONSTRAINT reporter_id_of_moderation_report FOREIGN KEY (`reporter_id`) REFERENCES login (id),
+  CONSTRAINT offender_id_of_moderation_report FOREIGN KEY (`offender_id`) REFERENCES login (id),
+  FOREIGN KEY (`last_moderator`) REFERENCES login (id)
+);
+
+CREATE TABLE IF NOT EXISTS reported_user (
```
axel1200: Is that needed? The information stored there seems to be redudant . https://en.wikipedia.org/wiki/Third_normal_form

migrations/V52__moderation_report.sql
```diff
@@ -0,0 +1,40 @@
+CREATE TABLE IF NOT EXISTS moderation_report (
+  `id`                 INT UNSIGNED     NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `create_time`        DATETIME        NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
axel1200: Should be last property (create and update time), it seems to be a convention in the db project

migrations/V52__moderation_report.sql
```diff
@@ -0,0 +1,40 @@
...
+  `update_time`        TIMESTAMP        NOT NULL DEFAULT CURRENT_TIMESTAMP  ON UPDATE CURRENT_TIMESTAMP,
+  `reporter_id`        MEDIUMINT UNSIGNED     NOT NULL,
+  `offender_id`        MEDIUMINT UNSIGNED     NOT NULL,
+  `report_description` LONGTEXT,
+ 
+  `game_id`            INT UNSIGNED     DEFAULT NULL
+  COMMENT 'If equal to NULL, the offense did not happen ingame',
+
+  `report_status`      VARCHAR(1)    NOT NULL
```
axel1200: Would be a lot better readable(for adminstrators to read table and people querring from the api) if you used mysql enums, well maybe there are not that performant. Maybe somebody can state his/her opinion on that

Rackover: I was discouraged to use Enums by (brutus or downlord I don't remember)

migrations/V52__moderation_report.sql
```diff
@@ -0,0 +1,40 @@
...
+  COMMENT 'Can be either [A]waiting, [P]rocessing, [C]ompleted, [D]iscarded',
+  
+  `game_incident_timecode`         VARCHAR(100)
+  COMMENT 'The timecode of the incident ingame, indicated by the user in their own terms',
+  
+  `moderator_notice`        LONGTEXT
+  COMMENT 'A public notice left by the moderator which will be addressed to the reporter once the report is marked as either COMPLETED or DISCARDED',
```
axel1200: How do we report that to him? Just interested, you already have an idea or will that be a future thing

Rackover: I will detail the full procedure when we will work on the API

migrations/V52__moderation_report.sql
axel1200: ok then just know we use them at other points, Maybe we can still write the whole word 🤔. But well downlord and brutus will know

Rackover: This has already been discussed with brutus though

norraxx: Enums/set are fine

axel1200: So you say we should use them

migrations/V52__moderation_report.sql
```diff
@@ -0,0 +1,40 @@
+CREATE TABLE IF NOT EXISTS moderation_report (
+  `id`                      INT UNSIGNED            NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `reporter_id`             MEDIUMINT UNSIGNED      NOT NULL,
+  `report_description`      LONGTEXT,
```
micheljung: Is a 16 MiB text (`MEDIUMTEXT`) or even 64 KiB (`TEXT`) not enough, do we really want to allow users adding up to 4 GiB text?
Same for other columns. Looks like we're allowing one report to be 12 GiB.

migrations/V52__moderation_report.sql
```diff
@@ -0,0 +1,40 @@
...
+ 
+  `game_id`                 INT UNSIGNED            DEFAULT NULL
+  COMMENT 'If equal to NULL, the offense did not happen ingame',
+
+  `report_status`           ENUM('Awaiting', 'Processing', 'Completed', 'Discarded')   NOT NULL
```
micheljung: If these values were UPPERCASE, they could very easily be mapped to code (and they also express themselves as enum values better, IMO). I'm not sure the mapping is as easy if the casing is different from in-code values.

axel1200:
Can we add this `ALTER TABLE ban ADD COLUMN (
    report_id INT UNSIGNED
    COMMENT 'The report based on which the ban was issued'
);
ALTER TABLE ban ADD CONSTRAINT FOREIGN KEY (report_id) references moderation_report(id);`

PR #166, opened 2018-09-11
Added last login time in users table

Rackover:
To use in conjunction with the python server to have a "last login" time


Rackover:
You want it to be Nullable ?

Rackover:
Or do I default it to Timestamp 0 ?

Brutus5000:
I would make it nullable

Rackover:
It is now

PR #172, opened 2018-09-23
Add telegram data

### LINKED ISSUE #170: Add telgram auth information
Add a colum to login table with the users telegram id

axel1200:
Fixes #170 
Would be nice if I could get a quick review that would allow me to test api changes on the test server.

### DIFF
--- a/faf/db/tables.py
+++ b/faf/db/tables.py
@@ -268,7 +268,6 @@ class Login(Base):
     salt = Column(String(16))
     email = Column(String(254, 'latin1_bin'), nullable=False)
     ip = Column(String(15, 'latin1_bin'))
-    steamid = Column(BigInteger)
     create_time = Column(DateTime, nullable=False, server_default=text("CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP"))
     update_time = Column(DateTime, nullable=False, server_default=text("'0000-00-00 00:00:00'"))
 
@@ -480,3 +479,10 @@ class VersionLobby(Base):
     id = Column(Integer, primary_key=True)
     file = Column(String(100))
     version = Column(String(100), nullable=False)
+
+class AccountAuthorization(Base):
+    __tablename__ = 'account_authorization'
+
+    id = Column(Integer, primary_key=True)
+    player_id = Column(Integer, nullable=false)
+    authorization_type = Column(Enum('steam','telegram'), nullable=False)
--- a/migrations/V60__telegram-auth.sql
+++ b/migrations/V60__telegram-auth.sql
@@ -1,0 +1,40 @@
+CREATE TABLE IF NOT EXISTS steam_user_id_hashes (
+  `id`                      INT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+  `hashed_telegram_user_id` varchar(64)                         NOT NULL UNIQUE,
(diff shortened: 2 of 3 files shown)


axel1200:
What if we decide to make double verification possible or we decide to not allow telegram anymore this modle would provide more flexibility.

micheljung:
I would prefer if we keep verification methods separate and if a user can have multiple verifications. If it was for me, I'd even display those in the user's profile.

**Downlord**
:heavy_check_mark: Verified by steam account
:heavy_check_mark: Verified by Telegram Passport
:heavy_check_mark: Well-known community member
:x: Verified by phone number

So I would work toward a flexible solution. Something like a generic table `account_verification` mc:1 `login` which has a enum-column `verification_type` and some value field(s) where we can enter telegram passport ID, steam ID, a mobile phone number's etc. either as hashed value or in plain, depending on what the requirements and/or use cases are.

This way we could support N verification methods while avoiding specialized columns and tables, making it easy to add or remove methods.

axel1200:
So I would need to change python server for this will somebody help me?

axel1200:
@Brutus5000 @micheljung how should i handle the steam id. Shell i still save which account has which steam id? or shall i go for hashing them all? And if yes have one table for all hashed authorization ids in one table.

axel1200:
I think @Brutus5000 idea was not not link any hashes or ids(form steam or telegram) to the user so they would need a separate table.

axel1200:
Look at the changes a made that is what i thought of

axel1200:
Would be nice if u rereviewed

axel1200:
Yeah I know I read that.

Geosearchef:
The plan was to consider adding UID OR Telegram (letting users choose) on top of Steam linking.

Brutus5000:
No it was not. This was definitely not common consensus.

axel1200:
@Geosearchef That is what most devs had in mind. But visonik was a different opinion on that.

axel1200:
why that ? Steam is more than sufficent

Brutus5000:
Telegram was an alternative solution to people who don't want to link steam for whatever reason. Goal: increase the potential playerbase (unification of the sets).
Adding telegram on top gives us an even smaller playerbase (the intersection of people who want to link to steam and use Telegram).

Geosearchef:
That's what people suggested during the debate in #general.

visionik:
For clarification, it's fine to add additional 'verification' methods if the community or devs want.  My point was just that we need to require steam link account verification and most importantly verification they own sc:fa via steam.  Another way of thinking about it is via Downlords comment above that users could have multiple verifications, I'm just saying steam link + game verify is the only mandatory verification.

axel1200:
Well i think neither geo nor i are in favor of making even stricter rules I think our goal was to stop decreasing number of ppl logging into faf, because we enforce steam link(Which you can clearly see in statistics we collect). The reason we enforced steam link was that we want to stop trolls like feather or super not that we wanted to verify the ppl own the game on steam. At least that was the reason @Rackover made the server PR from what I know. So geo and I thought of different mechanism on how to stop feather, super and co. What would be to telegram link or to make the UID system stricter. But steam link is more than sufficient for that. I find it sad we want to keep steam link  enforced but if that is legally necessary so it be.

PR #180, opened 2018-10-21
update FAF Develop description

PhilipJFryFAF:
changes the tool tip text on the faf develop game patch since the current one doesn't seem to encourage users to actually use it

`'Updated frequently for developer P2P testing'` -> `'Updated frequently for testing the upcoming game Patch'`


PhilipJFryFAF:
closing cause according to @Brutus5000 i need to make a new file with the same id to update stuff

Brutus5000:
This is "just" for initiating test databases. We don't need a migration for edit a descriptive test, I will do this by hand. But to keep it in sync, I merge this one as well.

PR #183, opened 2018-12-26
More coop map test data

Askaholic:
Add more coop maps (1 for each map type) and a delete statement to clear out old stuff


Brutus5000:
There seems to be a conflict with your prior PR. Please rebase.

Issue #185, opened 2019-01-04
Elide slows db


norraxx:
```
SELECT game0_.id AS id1_17_0_,
       gamereview2_.game_id AS game_id1_16_1_,
       game0_.endTime AS endTime2_17_0_,
       game0_.gameMod AS gameMod7_17_0_,
       game0_.host AS host8_17_0_,
       game0_.mapId AS mapId9_17_0_,
       game0_.gameName AS gameName3_17_0_,
       game0_.startTime AS startTim4_17_0_,
       game0_.validity AS validity5_17_0_,
       game0_.gameType AS gameType6_17_0_,
       gamereview2_.lower_bound AS lower_bo2_16_1_,
       gamereview2_.negative AS negative3_16_1_,
       gamereview2_.positive AS positive4_16_1_,
       gamereview2_.reviews AS reviews5_16_1_,
       gamereview2_.score AS score6_16_1_
FROM game_stats game0_
LEFT OUTER JOIN game_featuredMods featuredmo1_ ON game0_.gameMod=featuredmo1_.id
LEFT OUTER JOIN game_reviews_summary gamereview2_ ON game0_.id=gamereview2_.game_id
WHERE (featuredmo1_.id IN (6))
  AND (game0_.endTime IS NOT NULL)
  AND (game0_.validity IN (0))
  AND game0_.endTime>='2018-08-10 00:00:00.0'
  AND game0_.endTime<='2018-09-28 23:59:59.0'
ORDER BY game0_.startTime ASC
LIMIT 500
```
Speedup from 60s to 0.4 on test.

Brutus5000:
We fixed some of these issue. Some are unresolvable. Unless we have a new clearly indicated query I am closing this for now.

PR #186, opened 2019-01-04
Speedup game_stats all time recalculation query

norraxx:
Solution for #185


Brutus5000:
Please fix the travis build. You need a double underscore after the version.

PR #187, opened 2019-01-06
Add validity states for coop leaderboards

Askaholic:
Some additional reasons why a coop game may be considered 'unranked' and will not be inserted into the leaderboard.

Other projects that need updating:
  * [ ] ~~Faf Tools~~ <-- Is super outdated, and unused
  * [x] Python server https://github.com/FAForever/server/pull/388
  * [x] Python Client <-- Doesn't need update. It doesn't care about these values
  * [x] Java Commons https://github.com/FAForever/faf-java-commons/pull/23
  * [x] Java Api **<!-- Does NOT use Java Commons... Why?** https://github.com/FAForever/faf-java-api/pull/303
  * [x] Java Server **<!-- Does NOT use Java Commons... Why?** https://github.com/FAForever/faf-java-server/pull/111
  * [x] Moderator client <-- Uses Java Commons
  * [x] Java Client **<!-- Does NOT use Java Commons... Why?** https://github.com/FAForever/downlords-faf-client/pull/1093

### DIFF
--- a/migrations/V63__coop-validity-states.sql
+++ b/migrations/V63__coop-validity-states.sql
@@ -1,5 +1,5 @@
 INSERT INTO game_validity (id, message) VALUES
   (20, 'Civilians were revealed'),
-  (21, 'Difficulty was too low'),
+  (21, 'Difficulty was wrong'),
   (22, 'Expansion was disabled'),
   (23, 'Team spawn was not fixed');
--- a/migrations/V63__coop-validity-states.sql
+++ b/migrations/V63__coop-validity-states.sql
@@ -2,4 +2,5 @@ INSERT INTO game_validity (id, message) VALUES
   (20, 'Civilians were revealed'),
   (21, 'Difficulty was wrong'),
   (22, 'Expansion was disabled'),
-  (23, 'Team spawn was not fixed');
+  (23, 'Team spawn was not fixed'),
+  (24, 'Unranked for another reason');
--- a/migrations/V63__coop-validity-states.sql
+++ b/migrations/V63__coop-validity-states.sql
@@ -1,0 +1,5 @@
+INSERT INTO game_validity (id, message) VALUES
+  (20, 'Civilians were revealed'),
+  (21, 'Difficulty was too low'),
+  (22, 'Expansion was disabled'),
+  (23, 'Team spawn was not fixed');


Askaholic:
Maybe, but who is to say that these things aren't relevant for any other game mode? If you're game has one of these state, the your game also has the `mod: coop` tag so I don't see why that information needs to be duplicated. Maybe it's just a signal that there are some cases that we missed for what is currently allowed in ranked. Like if civilians are revealed, shouldn't that also unrank normal games?

Rackover:
And I agree with Askaholic here. If each change or addition to the ranked states needs a full upgrade of the faf-stack, there **must** be something we're doing terribly wrong.
The java server update is not that important for now, it can be done later.
If the work to do on the API is small maybe kubko can do it fast. Same for moderator client, or we can ask axel or geo. We don't lack java developers.
Python server is already done by askaholic (this PR is because of the python server, precisely).

Askaholic:
Also why would the clients need an update after a simple database insertion? It's just data...

Askaholic:
Looks like Java Server has a validity state `SERVER_SHUTDOWN` which doesn't exist in the DB.
https://github.com/FAForever/faf-java-server/blob/develop/faf-java-server-app/src/main/java/com/faforever/server/game/Validity.java

Rackover:
Even worse, java server has the wrong names for FFA and other invalidity states. If the point was to give some unrank conditions a better name, this is very poorly done - a change to the db and to the python server should have been done aswell.

micheljung:
> It should just be a string field or something

If we used a string field:
* It would take way more space than needed, because a whole string is stored for each game, and the string will be duplicated again and again
* We will no longer be able to interpret the data to answer questions like "What's the percentage of games invalidated due to desync?"
* We can no longer display a localized reason to the user

> there **must** be something we're doing terribly wrong.

* A less agressive, negative and more constructive phrase be: "Isn't there a better solution?"

> Also why would the clients need an update after a simple database insertion? It's just data...

Because we want the clients to be able to interpret the data, e.g. it could use symbols in the replay vault to denote the validity of a game. And it can only do that if it knows the possible values. Of course, a good client implementation would have a fallback for unknown values so that it doesn't break when something is added server-side

> Even worse, java server has the wrong names for FFA and other invalidity states. If the point was to give some unrank conditions a better name, this is very poorly done

It does not have the wrong names and it is not very poorly done because the name is an implementation detail of the server, it has no link to what names the python server or API uses, as in the database we don't store strings but numeric values. Please get your facts straight before stating things that aren't true and judging based on your misunderstanding.

Askaholic:
@micheljung True, I guess FAF is up in the 9million games now, that's a lot of extra strings to store :P.

Is there a good reason why only the moderator client is using the validity definitions from `faf-java-commons`? If the java server and api both used those it would be a lot easier to maintain consistency (and consequently be less of a barrier to development since you'd only have to perform 2-3 pull requests as opposed to 5-6).

Brutus5000:
@Askaholic The faf-java-commons classes define the data transfer objects. Inside api and server we have a different layer of the same classes with annotations for JPA/database access and access permissions.

The java client does not use it for historical reasons. faf-java-commons api classes were a side product of my work on the moderator client. At some point the java client should use them, but nobody had the time for such a refactoring (which brings no features or bugfixes and thus has no priority).

Brutus5000:
@Rackover Usually validation states don't fall from the sky. They were fix quite a long time and in my opinion we are doing something wrong if we add more of them month after month instead of thinking this through once.

Rackover:
@Brutus5000 We're adding them month after month because they should've been added before and they were not because the server stayed idle.
I've thought of something that would allow us to have less validity state : just make a big one for the rules being non default (NON_DEFAULT_RULES) and keep the others (DESYNC, MULTI_TEAM, etc...)

But do we really want that ? We would lose a lot of precision.
I think it's good we have the necessary amount of invalidity states. It's not abnormal to have to add new states when the ranking code of the python server gets fixed or becomes more accurate. In that case, it's because of the ranking of coop. So everything's cool.

Brutus5000:
Yes, I know that the case here is different.
I was not talking about the past, I am talking about the future. If you come up next week again and add 5 more I would be pissed because these managed releases take quite some time.

Askaholic:
The Agile mindset would be to not worry about what we might need in the future, and just add what we're actually using right now. Then again I agree with @Brutus5000  that it would be smart to think about what we want out of the validity states and take this opportunity to flesh them out fully if possible.

Right now a lot of the validity states are directly associated with a specific game option for example WRONG_VICTORY_CONDITION, NO_FOG_OF_WAR, CHEATS_ENABLED, etc. The question is, do we really want to have a unique validity associated with every possible game option? If yes, then I would just need someone to enumerate all game options (since I'm not familiar with the lobby code) and I can add a state for each one.

However, as @Rackover pointed out, maybe we don't want to do this because that's a lot of states, and it doesn't really gain a whole lot. I can see that it's useful to be able to query the number of desynced games, but the number of games that were unranked because they were using the wrong victory condition? Maybe not so useful. 

So maybe it would be better to have a more generic BAD_GAME_OPTION state that can be used if the game was invalidated because of a lobby setting. If we chose to do this the number of states goes down significantly (obviously we can't delete the old ones from the database, but we could deprecate them if needed). It could be something like this:

```Java
VALID,
TOO_MANY_DESYNCS,
BAD_GAME_OPTIONS,
BAD_MAP,
BAD_MOD,
COOP_UNRANKED,
MUTUAL_DRAW,
BAD_RESULT, // Contains UNKNOWN_RESULT, TOO_SHORT
BAD_TEAMS,  // Contains FREE_FOR_ALL, MULTI_TEAM, SINGLE_PLAYER
HAS_AI, // Maybe should be thrown into BAD_GAME_OPTIONS
SERVER_ERROR
```

So it comes down to the trade off of being able to tell which exact check unranked the game, and having to keep a more complex list of states in sync across 6 different repos.

Rackover:
We could also add OTHER_UNRANK as a default state so that we could deploy server changes without to wait on all the rest - and unknown states will just be dealt with as OTHER_UNRANK.

Askaholic:
@Brutus5000 Yea I can add an `OTHER_UNRANK`. 

The only problem I see with OTHER_UNRANK is that it will be an easier option than adding proper states, and so people will just opt to use it and not bother adding a proper state. So if the ranking system changes over time I can see OTHER_UNRANK being pervasive, and kindof taking over all the other states. 

For example the PR that I'm working on to fix the coop leaderboards. If OTHER_UNRANK had existed I would have just used that for everything. So in a way, **not** having it is forcing us to do better error reporting.

That's jut my gut feeling, but maybe I'm wrong. Obviously, you have been working on FAF way longer than me so I'll let you make the decision. But I will gladly do the work of adding it to all these 6 related PR's.

PR #196, opened 2019-07-04
Add featured mods for missions

### LINKED ISSUE #163: Add Support For Featured Mods (per mission).
Add an extra table or column to the FAF DB to support featured mods with different missions. We're currently waiting on this for Nomads Coop. The idea would be to set a mod for each mission (E.G. All existing coop maps would run with the original `init_coop` mod, and Nomads missions should run with `init_coop_n` mod). The files can be found on our organisation for the Nomads Coop Featured Mod. Once this has been completed, the client maintainers can fetch the relevant information and files from the DB and apply them before the game is launched.

Nomads Coop Featured Mod: https://github.com/FAForever/NCoop

1-alex98:
Fixes #163

### DIFF
--- a/migrations/V71__allow_featured_mod_per_misson.sql
+++ b/migrations/V71__allow_featured_mod_per_misson.sql
@@ -1,0 +1,19 @@
+CREATE TABLE coop_map_type(
+    `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
+    `name` VARCHAR(256) not null ,
+    `description` text,
+    `main_featured_id` tinyint(3) unsigned NOT NULL,
+    PRIMARY KEY (`id`),
+    CONSTRAINT `featured_mod_for_coop_map_type` FOREIGN KEY (`main_featured_id`) REFERENCES `game_featuredMods` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE TABLE coop_map_type_to_needed_featured_mods(
+    `coop_map_type_id` mediumint(8) unsigned NOT NULL,
+    `needed_featured_id` tinyint(3) unsigned NOT NULL,
+    CONSTRAINT `coop_map_type_id_for_needed_featured_mods` FOREIGN KEY (`coop_map_type_id`) REFERENCES `coop_map_type` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
+    CONSTRAINT `featured_mod_id_for_needed_featured_mods` FOREIGN KEY (`needed_featured_id`) REFERENCES `game_featuredMods` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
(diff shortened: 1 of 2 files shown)


Brutus5000 (changes requested):
Why do we need 2 additional tables? Just add a foreign key to coop_map with reference to the featured mod.

axel1200:
no we need both, for nomands coop we need to download and update coop featured mod  and nomands featured mod and then start nomands featured mod. And maybe we have some other coop that depends on several featured mods....

axel1200:
It seems like @Exotic-Retard needs to write what is needed... Are both coop and nomands neeed to launch nomads coop. And are special versions needed or will the newest version always we the correct one. Which init needs to be launched...?????

Exotic-Retard:
1.Newest version of nomads will always be the correct one
2.Nomads and coop is needed, to be run in one init, more specifically this init: https://github.com/FAForever/NomadMissions/blob/master/init_NomadsCoop.lua
3.the init should be defined by whatever featured mod is used, probably.
4.we could create a nomads coop featured mod for this purpose, which may be useful to have regardless.

axel1200:
Oder man checkt im modifizierten nomands init ob es eine coop mission ist und launcht das entsprechende init. Na ja ka aber es gibt meinem Schema eigentlich Recht. Außer Brutus will eine weitere featured mod anlegen, die beides kombiniert. Musst du wissen Brutus....

axel1200:
Completely forgot should write English

Or you check in the modified nomands init if it is a coop mission and launch the corresponding init. Well no idea but it actually shows why my scheme makes sense. Well only in case Brutus does not want to create another featured mod that combines both.You need to know Brutus....

Exotic-Retard:
`Or you check in the modified nomands init if it is a coop mission and launch the corresponding init.` im not sure this is possible inside init files specifically - a bunch of engine commands are not yet loaded, and i could not find a way to check what arguments the game is launched with at that stage. checking in the lobby is too late at that point since the files are already loaded. (and theres currently no known method of reloading them while inside the lobby, which would also require a lobby reload.)

axel1200:
Well my solution would also offer making a new mod for nomands coop with only the init file and nothing else. Then enter nomands and coop as secondary/needed featured mods and tada

axel1200:
I joint featured mod might be hard to update... At least thats what I fear

Brutus5000:
I don't think we can allow multi-featured mod dependencies. If I understand correctly then
- coop depends on faf
- nomads depends on faf
- nomads-coop depends on coop **and** nomads

If nomads coop now depends on coop and nomads we'd need to introduce some sort of dependency management to resolve order of loading etc. This is a complexity that we need to avoid at all cost.

axel1200:
ok so we try to make a combi featured mod...

axel1200:
We can write a py script that reads the db tables and joins them and adds the extra init. Maybe like this.

Brutus5000:
The first batch of tasks to achieve the goal of this issue is outlined in https://github.com/orgs/FAForever/projects/8

Issue #190, opened 2019-03-17
ladder_1v1 mod has to be specialcased when accessing updates_* tables


Wesmania:
The ladder1v1 mod does not have its own files and just uses the main FAF mod. It sucks a bit for code that relies on modnames to access the updates_* tables (and there's unfortunately no other way to do that). I guess one way to fix that would be to add views that just point to the FAF mod tables, or maybe just rework the multiple tables into a single table set? We could still expose views that access the single table set underneath for compatibility.

Brutus5000:
Afaik this is obsolete since matchmaker.

PR #197, opened 2019-07-11
Add tables needed for multiple ratings and multiple map pools

Askaholic:
Copying over tables from the SupComHub schema: https://gitlab.com/supcomhub/sh-core-db-migrations/

This will be needed for team matchmaking and it will also be able to address existing issues such as:
  * Separate queries for global_rating and ladder1v1_rating (global missing win_games column)
  * Will be able to use a ladder1v1 leaderboard instead of a ladder1v1 featured mod (#190).


migrations/V71__leaderboards_and_map_pools.sql
```diff
@@ -0,0 +1,47 @@
...
+    `update_time`   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE leaderboard_rating
+(
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    login_id       INT       NOT NULL REFERENCES login (id),
+    mean           FLOAT     NOT NULL,
+    deviation      FLOAT     NOT NULL,
+    rating         FLOAT     GENERATED ALWAYS AS (mean - 3 * deviation), -- TODO: is this column even necessary?
```
Askaholic: This was taken from the supcomhub schema, but does it even make sense for this to be in the database? Also I believe FAF currently calculates the rating as `mean - 2*deviation`.

Rackover: No its 3 in FAF
I dont think its immensely needed, but why not I guess

migrations/V71__leaderboards_and_map_pools.sql
```diff
@@ -0,0 +1,47 @@
...
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    game_player_stats_id    INT REFERENCES game_player_stats (id),
+    leaderboard_id          INT REFERENCES leaderboard (id),
+    rating_mean_before      FLOAT      NOT NULL,
+    rating_deviation_before FLOAT      NOT NULL,
+    rating_mean_after       FLOAT      NOT NULL,
+    rating_deviation_after  FLOAT      NOT NULL,
+    `create_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `update_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
```
norraxx: Add multi-column key on game player stats id and leaderboard id

migrations/V71__leaderboards_and_map_pools.sql
norraxx: What is default, stored or virtual? Maybe it should bee stored explicitly?

migrations/V71__leaderboards_and_map_pools.sql
```diff
@@ -0,0 +1,47 @@
...
+
+CREATE TABLE matchmaker_pool
+(
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    technical_name  VARCHAR(255) NOT NULL UNIQUE,
+    featured_mod_id INT          NOT NULL REFERENCES game_featuredMods (id),
+    leaderboard_id  INT          NOT NULL REFERENCES leaderboard (id),
+    name_key        VARCHAR(255) NOT NULL,
+    `create_time`   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `update_time`   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
```
norraxx: Remove backticks or add them everywhere ;-)

migrations/V71__leaderboards_and_map_pools.sql
Brutus5000: It helps querying in the API. If you sort and filter on computed attributes you run into some issues.
Also this allows us to use other rating systems than Trueskill - but that only makes sense if we offer it in the journal table as well.

migrations/V71__leaderboards_and_map_pools.sql
Brutus5000: The four rating columns are not nullable. Why? 
Right now null values before are an indicator for the player not having a rating and null values after are an indicator for the game not being rated for various reasons.
How would we do this here?

migrations/V71__leaderboards_and_map_pools.sql
Askaholic: Somehow I think trueskill is too deeply embedded in FAF that it's going to be changing any time soon. I suppose allowing other leaderboards to use different rating systems could be nice.

migrations/V71__leaderboards_and_map_pools.sql
Askaholic: I think `NOT NULL` is appropriate for the journal because there is nothing to write into the journal in both of the cases you described.

1. The player doesn't have a rating.
Then there also don't need to be any journal entries for that rating. If the player is new, then you could write the journal entry with before and after ratings both set to the same initial rating.
2. The game is unranked.
In this case the rating doesn't change which means there is no reason to create a journal entry.

Askaholic: Can you elaborate? Do you mean a multi column foreign key like this: https://stackoverflow.com/questions/953035/multiple-column-foreign-key-in-mysql, or a multi column primary key, or an index? What is the reasoning for this?

norraxx: Multi column index, for better search

Brutus5000: I don't agree on unranked games. Leaving out the record because it does not change anything makes unclear whether data is missing by error or intentionally left missing. Also suddenly you have less journal records than game players stats records. That would be very confusing.

And the "my rating did not update properly please investigate" is a fairly common support question.

Askaholic: Well if you want to journal a `no change` action then your "before" and "after" states will be the same. So if you want, you can insert an entry into the journal table with the before and after fields identical. However, it's going to generate a lot of data just to say "yea nothing happened to your rating this game", so to me it seems like it isn't necessary.

From my understanding of journals, you only write to the journal if you actually intend on making a change. If there's no change, then there's no journal entry. https://en.wikipedia.org/wiki/Transaction_log

Either way, I don't see an issue with having the fields be `NOT NULL`. I would think that writing a null value for `after_rating` columns would imply that the rating was deleted. And why would that ever happen?

Brutus5000: Is there a situation in the server where some peoples rating are changed and others are not? I remember on custom games you can leave within <no. of players> minutes without affecting your rating. Does this apply to all players in that case?
Basically, I want to be able to distinguish between "no rating change" due to the invalidity of a player result and "no rating change, because the system gained no additional insight".

Askaholic: As far as I can tell from the code (the game code is an absolute mess), ratings are either calculated for no players or all players depending on if the game is valid or not. The way the minimum time thing works is that if the game is too short it is marked as invalid meaning that ratings will not be calculated for that game. So currently you can look at the game_stats table to see if a game was valid or not.

Not sure if there are any plans for doing game validity on a per player basis. What would be the use case for that?

migrations/V72__leaderboards_and_map_pools.sql
```diff
@@ -0,0 +1,48 @@
...
+    name_key        VARCHAR(255) NOT NULL,
+    description_key VARCHAR(255) NOT NULL COMMENT "The leaderboard's i18n description key",
+    create_time   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    update_time   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE leaderboard_rating
+(
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    login_id       INT       NOT NULL REFERENCES login (id),
```
Brutus5000: Can we call it user_id instead? I hope to get rid or rename the login table at some point.

Askaholic: Might as well I guess. Other tables already use `player_id`, `user_id`, `idUser`, and `playerId`. Seems like `player_id` is more common. Maybe that's better?

PR #199, opened 2019-08-05
Add field to game_stats for storing in-game length

Askaholic:
With some changes across the replay server, API and client this should make vault displayed game lengths align perfectly with in game times. This will help remediate countless issues like this one:
https://github.com/FAForever/server/issues/346

That being said, I still would like the server game ending logic to get fixed so that we can have an accurate measurement of real time elapsed as well. For example, it would be interesting to browse the vault and compare in game lengths to real time elapsed. Although, I suspect that when people are browsing the vault, they care less about the real time elapsed than they do the actual game time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,6 @@ python: "3.6"
 services:
   - docker
 
-addons:
-  apt:
-    sources:
-    - mysql-5.7-trusty
-    packages:
-    - mysql-client
-
 before_script:
   - sudo /etc/init.d/mysql stop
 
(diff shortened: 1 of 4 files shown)


Brutus5000:
Actually, I would like to see an actual implementation using this before merging...

Askaholic:
@Brutus5000 You're in luck! https://github.com/FAForever/faf-aio-replayserver/pull/19

As far as the Java API/client is concerned I wouldn't even know where to begin. However, game time can be displayed easily in python like this:
```python
from datetime import timedelta

replay_ticks = 65530
print(timedelta(milliseconds=replay_ticks * 100))
# outputs: 1:49:13
```

PR #204, opened 2019-09-21
Added league / ladder week events

Rackover:
This way I dont have to store them in a loose json file at the root of the website, and we can also keep track of past ladder events and recalculate their scores

### DIFF
--- a/migrations/V77__league_events
+++ b/migrations/V77__league_events
@@ -1,0 +1,25 @@
+CREATE TABLE IF NOT EXISTS `league_event` (
+  `id`                      INT UNSIGNED           NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  
+  `start_time`              TIMESTAMP              NOT NULL,
+  `end_time`                TIMESTAMP              NOT NULL,
+  
+  `win_score`               FLOAT                  NOT NULL DEFAULT 1,
+  `draw_score`              FLOAT                  NOT NULL DEFAULT 0,
+  `loss_score`              FLOAT                  NOT NULL DEFAULT -1,
+  
+  `create_time`             TIMESTAMP              NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`             TIMESTAMP              NOT NULL DEFAULT CURRENT_TIMESTAMP  ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS `league_event_rating_category` (
+  `id`                      INT UNSIGNED           NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `league_event_id`         INT UNSIGNED           NOT NULL,
+  
+  `rating`                  INT                    NOT NULL,
+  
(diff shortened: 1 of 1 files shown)


Rackover:
Structure replicated from https://github.com/FAForever/website/blob/develop/configuration/leagues.json

migrations/V77__league_events
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE IF NOT EXISTS `league_event` (
+  `id`                      INT UNSIGNED           NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  
+  `start_time`              TIMESTAMP              NOT NULL,
+  `end_time`                TIMESTAMP              NOT NULL,
+
```
norraxx: Use datetime, not timestamp, datetime is more flexible, + timezone

migrations/V77__league_events
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE IF NOT EXISTS `league_event` (
+  `id`                      INT UNSIGNED           NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  
+  `start_time`              TIMESTAMP              NOT NULL,
+  `end_time`                TIMESTAMP              NOT NULL,
+  
+  `win_score`               FLOAT                  NOT NULL DEFAULT 1,
+  `draw_score`              FLOAT                  NOT NULL DEFAULT 0,
+  `loss_score`              FLOAT                  NOT NULL DEFAULT -1,
+
```
norraxx: What's the reason for float?

migrations/V77__league_events
```diff
@@ -0,0 +1,25 @@
...
+  
+  `create_time`             TIMESTAMP              NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `update_time`             TIMESTAMP              NOT NULL DEFAULT CURRENT_TIMESTAMP  ON UPDATE CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS `league_event_rating_category` (
+  `id`                      INT UNSIGNED           NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `league_event_id`         INT UNSIGNED           NOT NULL,
+  
+  `rating`                  INT                    NOT NULL,
```
norraxx: Can be rating signed? Or it's always unsigned?

migrations/V77__league_events
Rackover: In case someone wants half a point for draws for example. More flexibility

migrations/V77__league_events
Rackover: Rating brackets can be negative since its going to be computed rating

migrations/V77__league_events
Rackover: But in the whole rest of the DB createTime and StartTime are always timestamp... Shouldnt I stay consistent?

Brutus5000:
Use the tables you introduced a year ago.
https://github.com/FAForever/db/blob/develop/migrations/V56__leagues.sql

Rackover:
I'm very, very smart.

PR #205, opened 2019-09-23
Split permission table names into technical and translatable name

Brutus5000:


migrations/V77__permission_refinement.sql
```diff
@@ -0,0 +1,9 @@
+alter table group_permission
+    change name technical_name varchar(100) not null comment 'hardcoded names to be used in code';
+alter table group_permission
+    add name_key varchar(100) not null comment 'name key of permission in translation table';
+
+alter table user_group
+    change name technical_name varchar(100) not null comment 'hardcoded names to be used in code';
+alter table user_group
+    add name_key varchar(100) not null comment 'name key of user group in translation table';
```
norraxx: Will we query by this columns? If yes, will this columns be in tuples? If yes, add proper indexing.

Brutus5000: Right now I don't see a use case. We load the user into memory with all groups and permission and decide on that.

PR #206, opened 2019-09-23
Create V78__open_clans

Rackover:
Clans can have an "open" parameter which will allow anyone to join them without invite links.
(By default it's obviously to false)


Askaholic:
Nice. I think that's a good idea. 

It might be also be a good idea to add the ability for the owner of an open clan to kick or ban people from the clan. This is obviously more annoying to implement, and might not be necessary, but I think it's worth thinking about. Clan owners may want to make joining their clan easy, but still have the ability to vet people if they are somehow causing trouble.

Rackover:
> Nice. I think that's a good idea.
> 
> It might be also be a good idea to add the ability for the owner of an open clan to kick or ban people from the clan. This is obviously more annoying to implement, and might not be necessary, but I think it's worth thinking about. Clan owners may want to make joining their clan easy, but still have the ability to vet people if they are somehow causing trouble.

Yes, I agree. That requires a lot more of implementation though, so let's first see how this goes. If a user is repeatedly kicked and rejoined the clan leader will fill a report and we moderators can follow up on it. A clan-specific banning feature may not be required for the moment.

PR #208, opened 2019-11-26
Add column for game outcome

Askaholic:
As requested here: https://github.com/FAForever/server/pull/495#issuecomment-533682604.

I discussed this with @Wesmania earlier, and it would be quite easy to add on the server side.


Wesmania:
It's pretty easy to additionally split the UNDETERMINED value to distinguish no results and conflicting results. Would it be useful?

Askaholic:
Maybe. If there is no down side, might as well do it I guess.

migrations/V80__store_game_outcomes.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE game_player_stats
+  ADD COLUMN outcome ENUM(
+    'VICTORY', 'DEFEAT', 'DRAW', 'MUTUAL_DRAW', 'UNKNOWN_RESULT', 'CONFLICTING_RESULT'
+  ) default 'UNKNOWN_RESULT';
```
norraxx: Imho column can be named "result" and in enum values no need to have "_RESULT"

Askaholic: Yea, I agree. Changed

Askaholic:
The reason I haven't done that is because the field is supposed to reflect how the server rated the game, and that is dependent on the server implementation at the time. Also I'm not sure how relevant the information is for past games. Will anyone actually care?

But I could have  a look at the git blame and see how long the current code has been running on the server and at least compute the outcomes for those games.

norraxx:
fyi, don't make big update values in whole table. we should do such things in sort of maintenance tasks, in other words: run update query, x times with limit like 1000.

Askaholic:
Yea IDK if it's worth it. This is going to be a nasty query. For each game you need to sum team 2's points and team 3's points separately, then you need to set the results for each player depending on what team they were on and which team had the higher score. 

I'd much rather just right that as a python script instead of trying to do all that logic in a sql query directly. But idk how long that would take to run. Probably ages

Brutus5000:
Python script would be fine for me as well.

PR #213, opened 2019-12-29
Random IRC password

### LINKED ISSUE #212: Add random irc password for every user

1-alex98:


Brutus5000:
I do not understand what is happening here and what the intentions are.

axel1200:
@Brutus5000 Read the server PR comments or maybe join me on discord and I explain u 😛

axel1200:
So the irc password can no longer be derivted from the password because we don't store the password any longer so know we need a secrete/random password for irc for every user. @Askaholic just generates random passwords everytime a user logs in that is wrong tho because obviously the user won't be able to log in the second time if the password is different... So it needs to be a random password but also it needs to be the same every time. So what do we do? Simple we store it in the db alongside the user and send it to the user when he logs in.

1-alex98:
Do not think this is a good idea any longer

PR #214, opened 2020-03-15
TMM: Add tables for map pools

Askaholic:
For team matchmaking we need to be able to support multiple map pools. This PR adds the ability to create any number of map pools, and set one map pool as the active one for each matchmaker pool. 

Things I'd like some feedback on:
- Should we add a description to `map_pool`? I'm imagining that the name/description of map pools will only serve as a tool to help the admins remember what the map pool was designed for and will never be shown to regular users.
- Should `map_pools_maps` many-to-many table have create/update time? I'm thinking that's probably too much data which will have little value.

### DIFF
--- a/migrations/V82__map_pools.sql
+++ b/migrations/V82__map_pools.sql
@@ -10,9 +10,9 @@ COMMENT="A collection of maps, used by a matchmaker_pool for map selection";
 CREATE TABLE map_pool_maps
 (
     map_pool_id   INT NOT NULL REFERENCES map_pool (id),
-    map_id        INT NOT NULL REFERENCES map_version (id),
-    PRIMARY KEY (map_pool_id, map_id),
-    INDEX (map_id, map_pool_id)
+    map_version_id        INT NOT NULL REFERENCES map_version (id),
+    PRIMARY KEY (map_pool_id, map_version_id),
+    INDEX (map_version_id, map_pool_id)
 );
 
 CREATE TABLE matchmaker_pool_map_pools
--- a/migrations/V82__map_pools.sql
+++ b/migrations/V82__map_pools.sql
@@ -1,3 +1,6 @@
+ALTER TABLE matchmaker_pool RENAME TO matchmaker_queue;
+ALTER TABLE matchmaker_queue COMMENT="A matchmaker queue specifying which featured mod will be played, which map pools will be drawn from, and which leaderboard will be used to look up and update a player''s rating.";
+
 CREATE TABLE map_pool
 (
     id                  INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
@@ -15,12 +18,12 @@ CREATE TABLE map_pool_map_version
     INDEX (map_version_id, map_pool_id)
 );
 
-CREATE TABLE matchmaker_pool_map_pool
+CREATE TABLE matchmaker_queue_map_pool
 (
-    matchmaker_pool_id  INT REFERENCES matchmaker_pool (id),
+    matchmaker_queue_id  INT REFERENCES matchmaker_queue (id),
     map_pool_id         INT NOT NULL REFERENCES map_pool (id),
     min_rating          INT,
     max_rating          INT,
-    PRIMARY KEY (matchmaker_pool_id, map_pool_id),
-    INDEX (map_pool_id, matchmaker_pool_id)
+    PRIMARY KEY (matchmaker_queue_id, map_pool_id),
+    INDEX (map_pool_id, matchmaker_queue_id)
 );
--- a/migrations/V82__map_pools.sql
+++ b/migrations/V82__map_pools.sql
@@ -2,9 +2,6 @@ CREATE TABLE map_pool
 (
     id                  INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
     name                VARCHAR(255) NOT NULL,
-    matchmaker_pool_id  INT REFERENCES matchmaker_pool (id),
-    min_rating          INT,
-    max_rating          INT,
     create_time         TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
     update_time         TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
 ) ENGINE=InnoDB
@@ -16,5 +13,14 @@ CREATE TABLE map_pool_maps
     map_id        INT NOT NULL REFERENCES map_version (id),
     PRIMARY KEY (map_pool_id, map_id),
     INDEX (map_id, map_pool_id)
-)
-COMMENT="Map pool many-to-many table";
+);
+
+CREATE TABLE matchmaker_pool_map_pools
+(
+    matchmaker_pool_id  INT REFERENCES matchmaker_pool (id),
+    map_pool_id         INT NOT NULL REFERENCES map_pool (id),
+    min_rating          INT,
+    max_rating          INT,
+    PRIMARY KEY (matchmaker_pool_id, map_pool_id),
+    INDEX (map_pool_id, matchmaker_pool_id)
+);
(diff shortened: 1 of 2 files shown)


Geosearchef:
That's not the way we want to implement map pools. As I told you, can we please first talk about this?

migrations/V82__map_pools.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE map_pool
+(
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    name          VARCHAR(255) NOT NULL,
+    create_time   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    update_time   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+) COMMENT="A collection of maps, used by a matchmaker for map selection";
```
norraxx: engine=innodb

migrations/V82__map_pools.sql
```diff
@@ -0,0 +1,17 @@
...
+
+CREATE TABLE map_pool_maps
+(
+    map_pool_id   INT REFERENCES map_pool (id),
+    map_id        INT REFERENCES map_version (id),
+    PRIMARY KEY (map_pool_id, map_id)
```
norraxx: add `key (map_id, map_pool_id)`

migrations/V82__map_pools.sql
norraxx: why do we have such `create_time` and `update_time`? If we will filter/search by this values, we should add indexes now. Otherwise it will be pain.

migrations/V82__map_pools.sql
Askaholic: Explain why

norraxx: Reason is joining.
MySQL during query execution creates execution plan, it's the left handed tree (such data structure), that is made by query optimizer.
Query optimizer can swap tables as much as it want, unless `STRAIGHT JOIN` is in query.
And now imagine 3 tables:
`a`, (our `map_pool_maps`) and table `c` and then:
`a` join `map_pool_maps` on `map_pool_id` column and then table `c` on column `map_id`, but WHAT IF there is filter (where condition) on table `c`, like `c`.`id` = 1?
Then execution plan should be something like:
`c` join `map_pool_maps` on `map_id` and then join `a`.

And such execution plan WILL be created, BUT it will be ineffective.

And now the important question: `How MySQL uses coverage indexes?`
It uses them from left to right. First you will give columns for joins, then for filters, then for result.

So as you see, with the second case from `c` to `a` join, there will be no reason to:
1. open main table
2. do the full table scan
3. multiply results and do big temporary table in memory

Just open index, and use it for first and second join. With primary index it will work same (from `a` to `c`), way but you defined it explicit.

Did I describe it enough?

migrations/V82__map_pools.sql
Askaholic: I wasn't planning on searching by them. They would be just for historical purposes.

migrations/V82__map_pools.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE map_pool
+(
+    id                  INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    name                VARCHAR(255) NOT NULL,
+    matchmaker_pool_id  INT REFERENCES matchmaker_pool (id),
```
Askaholic: Right now this is set up as a many-to-one relationship to matchmaker_pool. Do we instead want a many-to-many relationship? Is it conceivable that we would want to set up a single map pool that is used by multiple queues? For instance a casual and a tryhard version of the same queue.

Geosearchef: I think a many-to-many relationship is a good idea. It will then apply the same rating range to both queues though.

Geosearchef (approved):
looks good to me

migrations/V82__map_pools.sql
Askaholic: We could also add the rating ranges into the many-to-many table. Not sure if we'd want that, but it probably doesn't hurt to have the flexibility right? It will possibly mean extra data in the database, but I don't think it will be a big deal since we probably wont ever have more than a handful of queues, and eve over long periods of time probably only hundreds of map pools.

migrations/V82__map_pools.sql
```diff
@@ -0,0 +1,26 @@
...
+    PRIMARY KEY (map_pool_id, map_id),
+    INDEX (map_id, map_pool_id)
+);
+
+CREATE TABLE matchmaker_pool_map_pools
+(
+    matchmaker_pool_id  INT REFERENCES matchmaker_pool (id),
+    map_pool_id         INT NOT NULL REFERENCES map_pool (id),
+    min_rating          INT,
+    max_rating          INT,
```
norraxx, line 26: I don't see the logic behind that two columns  (min/max rating), but i think there will be some sort of computation with it, and i suggest to make add them to indexes, primary will have them automatically, but secondary might have them too.
Reason: every secondary index has pointer to primary index, and when we search from secondary index and asking for values that are not in secondary index primary index is opened and searched. For such small table - there is no reason to open primary index.

Askaholic: I don't think we need to add an index to these columns. We will never fetch a map pool based on rating. These tables are assumed to be changing very infrequently so we will just fetch them once per 10 minutes and cache the results in memory.

1-alex98:
@Brutus5000 merge or not merge?

1-alex98:
@Askaholic try to use the original repo and not a fork now we can not see if the CI runs through fro rxample

1-alex98:
@Brutus5000 I think you are the only one being allowed to merge on the db repo

migrations/V82__map_pools.sql
```diff
@@ -0,0 +1,26 @@
...
+    name                VARCHAR(255) NOT NULL COMMENT "Only to help admins with organization. Don't show this to the user.",
+    create_time         TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    update_time         TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+) ENGINE=InnoDB
+COMMENT="A collection of maps, used by a matchmaker_pool for map selection";
+
+CREATE TABLE map_pool_maps
+(
+    map_pool_id   INT NOT NULL REFERENCES map_pool (id),
+    map_id        INT NOT NULL REFERENCES map_version (id),
```
Brutus5000: If it references a map version, it should also be called map_version_id

migrations/V82__map_pools.sql
```diff
@@ -0,0 +1,26 @@
...
+    PRIMARY KEY (map_pool_id, map_id),
+    INDEX (map_id, map_pool_id)
+);
+
+CREATE TABLE matchmaker_pool_map_pools
```
Brutus5000: I do get the intention of the table, but the naming is unpleasant.
Going with existing namings we would have something like matchmaker_pool_assignments.

Brutus5000 (changes requested):

migrations/V82__map_pools.sql
Askaholic: Hmm, I agree its not a nice name, the only problem with `matchmaker_pool_assignments` is that it doesn't capture the bidirectional relationship. Like `map_pool_maps` would become `map_pool_assignment` which doesn't make nearly as much sense.

Brutus5000: Maybe matchmaker_map_subpools ?

Askaholic: Or `matchmaker_map_pools`. I could rename `matchmaker_pool` too since it isnt used yet. I'm not sure why it even has `pool` in the name

Brutus5000 (approved):

PR #216, opened 2020-03-28
Cleans up invalid map ids for old games

### LINKED ISSUE #215: Some old games use 0 as map Id istead of null and others are simply invalid
Add a forgein key constraint in the game_stats table and clean all invalid map ids beforehand(set them to null)

1-alex98:
Fixes #215

### DIFF
--- a/migrations/V83__invalid_map_ids_in_games.sql
+++ b/migrations/V83__invalid_map_ids_in_games.sql
@@ -1,0 +1,2 @@
+UPDATE game_stats set mapId = null where mapId = 0 or mapId IS NOT NULL AND NOT EXISTS(SELECT * FROM map_version WHERE id = game_stats.mapId);
+ALTER TABLE game_stats ADD CONSTRAINT mapid_fkey FOREIGN KEY game_stats(mapId) references map_version(id);


1-alex98:
tested on test server

Brutus5000:
Not every player plays with a map uploaded to the server. Null values are acceptable.

Brutus5000:
I obviously misread the the description

migrations/V82__invalid_map_ids_in_games.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE game_stats set mapId = null where mapId =0 or not exists(SELECT * from map_version where id= game_stats.mapId);
```
norraxx: IMHO: Second condition in `or` should be checked on not null
`or mapId != NULL AND NOT EXISTS(SELECT * FROM map_version WHERE id = game_stats.mapId)`

Reason: speedup.

1-alex98: really I assumed that psql is inteligent enought to evaluate the first part before the second and stopping otherwise

1-alex98: What u suggest it logically redundant

1-alex98:
@Brutus5000 thx for the approve but I got no rights to merge

Brutus5000:
I did not approve this, I just reopened it. I need to know if null values are still possible in game with the foreign key. Not sure right now.

1-alex98:
they are

1-alex98:
Oh i confused the grenn light with the reopen with approve green symbol

migrations/V82__invalid_map_ids_in_games.sql
norraxx: Psql? Since when we use postgres? We use mysql.
It is not redundant.

Mysql will lock all rows, even is they don't need to be locked.

1-alex98: as u pleasse

1-alex98: better?

norraxx: :heart:

Brutus5000:
@axel1200 Please move this to v83

1-alex98:
@Brutus5000 done

Brutus5000 (approved):

PR #217, opened 2020-04-11
Add leaderboard entries for global and ladder_1v1.

cleborys:
Populate `leaderboard` with rating types for global and ladder 1v1, see #197.

Necessary for https://github.com/FAForever/server/pull/553 / https://github.com/FAForever/server/issues/534.
 
I hope I am using `name_key` and `description_key` as intended.


migrations/V84_add_leaderboard_entries.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO leaderboard (id, technical_name, name_key, description_key) VALUES
+  (1, 'global', 'leaderboard.global.name', 'leaderboard.global.description'),
+  (2, 'ladder_1v1', 'leaderboard.ladder_1v1.name', 'leaderboard.ladder_1v1.description'),
+;
+
+INSERT INTO messages (key, language, region, value) VALUES
```
norraxx: BTW: where do we use localization?

norraxx:
For me seems good

migrations/V84_add_leaderboard_entries.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO leaderboard (id, technical_name, name_key, description_key) VALUES
+  (1, 'global', 'leaderboard.global.name', 'leaderboard.global.description'),
```
Askaholic: I'm not entirely sure how the i18n keys are supposed to work. I didn't even know there was a `messages` table, I assumed these keys would map to a file in the client somewhere. Maybe @Brutus5000 knows.

Brutus5000: https://github.com/FAForever/db/blob/develop/migrations/V1__base.sql#L1053

But it's not that well maintained.

Brutus5000 (changes requested):
The file needs a double underscore after the version number.

Brutus5000 (approved):

PR #219, opened 2020-04-18
Changing achievement statistics view to table

norraxx:


bukajsytlos (commented):
lgtm although not sure if 1 min is enough

norraxx:
It's more than enough, it will happen only once per whole server, shouldn't create load.
It's global user statistics count.

Askaholic:
What are achievement_statistics even used for? Just to know how many people have unlocked which achievements?

norraxx:
What I know, user profile -> statistics, and from chat -> right click on user -> profile -> statistics.

bukajsytlos (approved):

Brutus5000:
In the API the statistics are joined into the achievement entity (https://api.faforever.com/data/achievement).

That's also the reason why the server got overloaded when we started pushing achievements. Each time the lobby server pushes (at least once per second) the API loads the achievement and performs the statistics view query which takes 2s on a server with minimum load. (and went up to 60 seconds under heavy load)

Brutus5000 (approved):

PR #220, opened 2020-04-20
Database speed up getting rid of views

norraxx:
Another wave to get rid of views.

- map_reviews_summary
- ladder1v1_rating_distribution
- global_rating_distribution
- mod_review_summary

Depends on https://github.com/FAForever/db/pull/219

PS: To be continued.

### DIFF
--- a/migrations/V89__global_rating_distribution_view_to_event.sql
+++ b/migrations/V89__global_rating_distribution_view_to_event.sql
@@ -8,20 +8,10 @@ CREATE TABLE IF NOT EXISTS `global_rating_distribution` (
 ALTER TABLE `global_rating`
 ADD INDEX `global_rating_idx_for_rating_distribution` (`is_active`, `numGames`, `mean`, `deviation`);
 
-INSERT INTO global_rating_distribution
-SELECT `subq`.`rating`, COUNT(*)
-FROM (
-  SELECT floor((`global_rating`.`mean` - 3 * `global_rating`.`deviation`) / 100) * 100 AS `rating`
-  FROM `global_rating`
-  WHERE `global_rating`.`is_active` = 1 AND `global_rating`.`numGames` > 0
-) AS `subq`
-GROUP BY `subq`.`rating`
-ORDER BY `subq`.`rating`;
-
 DELIMITER $$
 
 CREATE EVENT IF NOT EXISTS `global_rating_distribution`
-ON SCHEDULE EVERY 1 HOUR
+ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
 COMMENT 'Recompute the global_rating_distribution table every hour'
 DO
     BEGIN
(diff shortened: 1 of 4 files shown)


migrations/V87__map_reviews_summary_view_to_event.sql
```diff
@@ -0,0 +1,78 @@
...
+  JOIN `map_version` on `map_version`.`id` = `summary`.`map_version_id`
+)
+GROUP BY `map_version`.`map_id`;
+
+
+DELIMITER $$
+
+CREATE EVENT IF NOT EXISTS `map_reviews_summary`
+ON SCHEDULE EVERY 10 MINUTE
+COMMENT 'Recompute the achievement_statistics table every minute'
```
bukajsytlos: copy paste typo

migrations/V89__global_rating_distribution_view_to_event.sql
```diff
@@ -0,0 +1,43 @@
...
+  WHERE `global_rating`.`is_active` = 1 AND `global_rating`.`numGames` > 0
+) AS `subq`
+GROUP BY `subq`.`rating`
+ORDER BY `subq`.`rating`;
+
+DELIMITER $$
+
+CREATE EVENT IF NOT EXISTS `global_rating_distribution`
+ON SCHEDULE EVERY 1 HOUR
+COMMENT 'Recompute the achievement_statistics table every minute'
```
bukajsytlos: same here

migrations/V87__map_reviews_summary_view_to_event.sql
norraxx: Well, it's the classic programming methodology `copy-paste kung-fu`.
You're reusing same code that works :joy_cat:. I'll fix it in the evening.

norraxx:
Another thing, that comes into my mind:
IF (with big letters) it is necessary to recompute this tables immediately (let's say that they're materialized views) i could extract event bodies into the functions, that will be avalilable for call like:
`select recompute_mod_reviews_summary();`

migrations/V89__global_rating_distribution_view_to_event.sql
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE EVENT IF NOT EXISTS `global_rating_distribution`
+ON SCHEDULE EVERY 1 HOUR
+COMMENT 'Recompute the global_rating_distribution table every hour'
+DO
+    BEGIN
+      START TRANSACTION;
+      DELETE FROM global_rating_distribution;
+
+      INSERT INTO global_rating_distribution
```
drahoja9, line 21: No need to repeat this `INSERT` once again, right?

norraxx: First insert will push data right during migration, to not to have gap, when data are not available.
second one is in event, which will delete everything and refresh the content of the table.

drahoja9: What about adding `STARTS CURRENT_TIMESTAMP` after `ON SCHEDULE EVERY 1 HOUR` (line 24)? That should insert data right after the event is created and replace the redundant `INSERT` statement above.

norraxx: done

PR #221, opened 2020-04-26
Add join_channel permission

Askaholic:


Brutus5000 (approved):
We'll need to adapt the version number (most probably v88)

Askaholic:
Yep. I'll do that

PR #223, opened 2020-04-27
Only consider global bans for lobby server bans

### LINKED ISSUE #222: Vault bans are considered as global bans
The view `lobby_ban` only ignores chat bans and as a result the lobby server takes a vault ban as overall ban, thus preventing the user from login

Brutus5000:
(fixes #222)


Brutus5000:
Already deployed on the server, waiting for positive test feedback.

PR #224, opened 2020-05-13
WIP: Add modifications for matchmaking

Askaholic:
Adding a nullable reference to `matchmaker_queue.id` in the `game_stats` table should give us a cleaner way of distinguishing between custom/ladder games without the need for a fake featured_mod.  We could even go through and link up all previous ladder games (including old games that were played on the `faf` featured mod) to make it easier to query stats on ladder games.

# TODOs
- [ ] Set `matchmaker_queue_id` for all existing ladder games? Maybe make it a separate migration so we can ensure that we don't miss any during transition.
- [ ] Should we add the 2v2 matchmaker queue and leaderboard entries in a migration? Or insert that data by some other means?


migrations/V89__matchmaker_enhancements.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE matchmaker_queue
+  ADD COLUMN team_size UNSIGNED TINYINT NOT NULL DEFAULT 1,
+  ADD COLUMN enabled BOOLEAN NOT NULL DEFAULT TRUE
```
norraxx: Will be there any queries with filtering over this columns?

Askaholic: Yea I think so

migrations/V93__matchmaker_enhancements.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE matchmaker_queue
+  ADD COLUMN team_size TINYINT UNSIGNED NOT NULL DEFAULT 1,
+  ADD COLUMN enabled BOOLEAN NOT NULL DEFAULT TRUE
+;
+
+ALTER TABLE game_stats
+  ADD COLUMN matchmaker_queue_id INT UNSIGNED REFERENCES matchmaker_queue (id)
```
Brutus5000: Wrong decleration of the foreign key constraint. See [manual](https://dev.mysql.com/doc/refman/5.6/en/create-table.html)
MySQL parses but ignores “inline REFERENCES specifications” (as defined in the SQL standard) where the references are defined as part of the column specification.
MySQL accepts REFERENCES clauses only when specified as part of a separate FOREIGN KEY specification.

Brutus5000: `game_stats` is the largest table we have with > 20 million records, so we really need to think about changes carefully.
a) When is this populated? Only for ladder and TMM? So only a small portion of the games at all? Then maybe an additional m:n table should be used.
b) Do we really need an int? A smallint should be sufficient. (The same should happen in leaderboard_* tables by the way)

Brutus5000 (changes requested):

migrations/V93__matchmaker_enhancements.sql
norraxx: :see_no_evil:

Askaholic: OK I think I'll do this in a follow up PR since I don't want to think about it right now :P

Askaholic: Wow I'm surprised I never heard about this in my databases course... We specifically used mysql for the entire relational db part of the course.

Brutus5000: norraxx and I encountered this on accident just 1 day before your PR as I did not find the index in my db tool :slightly_smiling_face:

Brutus5000 (approved):

PR #225, opened 2020-05-25
Add a login log to discover login attack attempts

Brutus5000:
We noticed a few hundred failed login attempts in a very short time which looks like a brute force password attack.

In order to detect and throttle such attempts, we need to log failed logins a) per ip and b) per user id. So we can see a) (distributed) attacks on a single account and b) single attacks across random accounts.

The idea is to store successful logins as well as failed ones. The api has to clear up the old attempts.

Logic could be as following:
A successful attempt on one account deletes older successfull logins on this account (aka only 1 successfull login attempt per account is stored). Failed login attempts have to be cleared out by the user (or are auto-cleared if the failed attempt count from the same ip is below a certain threshold)

### DIFF
--- a/migrations/V94__login_log.sql
+++ b/migrations/V94__login_log.sql
@@ -1,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS `login_log` (
+    `id` INT UNSIGNED NOT NULL UNIQUE AUTO_INCREMENT COMMENT 'ID for API/Elide compatibility',
+    `login_id` MEDIUMINT UNSIGNED NOT NULL,
+    `ip` VARCHAR(45) NOT NULL,
+    `attempts` MEDIUMINT UNSIGNED NOT NULL DEFAULT 1,
+    `success` BOOLEAN NOT NULL,
+    `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+    PRIMARY KEY (`success`,`ip`,`login_id`),
+    CONSTRAINT FOREIGN KEY (`login_id`) REFERENCES `login` (`id`)
+) ENGINE=InnoDB COMMENT='Log of logins to discover login attack attempts. Has to be cleaned up by the services.';


migrations/V93__login_log.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS `login_log` (
+    `id` MEDIUMINT UNSIGNED PRIMARY KEY AUTO_INCREMENT,
+    `login_id` MEDIUMINT UNSIGNED NOT NULL REFERENCES `login` (`id`),
+    `ip` VARCHAR(15) NOT NULL,
+    `attempts` MEDIUMINT UNSIGNED,
+    `success` BOOLEAN NOT NULL,
+    `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
```
Askaholic: Do we actually need an update time for this? I feel like this table is going to fill up very quickly so the less data the better.

norraxx: Wanna have create or update time?

migrations/V93__login_log.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS `login_log` (
+    `id` MEDIUMINT UNSIGNED PRIMARY KEY AUTO_INCREMENT,
```
norraxx: Primary key can be login_id, no need for column `id`

migrations/V93__login_log.sql
```diff
@@ -0,0 +1,9 @@
...
+    `login_id` MEDIUMINT UNSIGNED NOT NULL REFERENCES `login` (`id`),
+    `ip` VARCHAR(15) NOT NULL,
```
norraxx: https://stackoverflow.com/questions/6964143/storing-ipv6-addresses-in-mysql

I want us to have ip as binary stuff, and prepared for ipv6. + Use string to ip functions

norraxx: Also we should consider, how we want to work with that table. Do we want to search trough IPs?

migrations/V93__login_log.sql
```diff
@@ -0,0 +1,9 @@
...
+    `attempts` MEDIUMINT UNSIGNED,
+    `success` BOOLEAN NOT NULL,
+    `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+) ENGINE=InnoDB COMMENT='Log of logins to discover login attack attempts. Has to be cleaned up by the services.';
```
norraxx: What is workflow behind that table? What do we need that table for? Lookup table for one login_id? Or there is more job to do?
From my point of view we want to add indexes on this table. For me there are important values such as attempts and success. Also we could index the IP, which is more important thing. If there is bruteforce attack on multiple accounts from one ip, we can handle it better.

If botnet is attackingultiple accounts, we must search first by ip, then by login.

What i'm offering: 

One table as coverage index Index on: ip, login, createorupdatetime, attempts. (Everything can be as primary key, for faster lookup), and lesser redundancy.

Last success login is in login table already.

IP on first place, to select like 10 rows, if there is any suspicious activity, then login + timestamps. That's the filtering at the end is attack attempts count.

PS: writting from mobile with big fingers is really hard :-D

migrations/V93__login_log.sql
norraxx: When I'm thinking more, as pimary key should be ip + login_id.
If brute force attack is on one account, user won't be affected.

norraxx (commented):
Sorry for spam :see_no_evil:

migrations/V93__login_log.sql
Brutus5000: Yes we need an update time. The attempts counter will increase with every failed login from the same ip and so it will update the update_time. Therefore will know how many attempts per minute we have by checking (update_time - create_time) / attempts.

migrations/V93__login_log.sql
Brutus5000: Good point.
We'll want to aggregate by ip and aggregate by login id.

migrations/V93__login_log.sql
Brutus5000: As described in the PR text:

- aggregate by ip to see attacks from same ip
- aggregate by login to see attacks on a single user account (even if distributed)

norraxx: I was thinking a little bit more...
And I have 2 more things:

1. We chould have in memory table, because we don't need transactions, acid, and whole stuff that InnoDB offers, we don't even need `b+tree` indexes, that are not available for in memory tables, aggregations by hashes are brutal fast.

2. Well, we don't need database, we could have sqlite or even dict on lobby server. :angel:

Brutus5000: I want this to be persisted beyond an api restart. And why would I use sqlite if I have MySQL running already?

norraxx: How about to be "persisted" in MySQL in memory table? :-)

norraxx: Well, I'll spam you a bit with ideas on zulip :wink:

Brutus5000: I see no advantage from this. It's not performance critical at all.

migrations/V93__login_log.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS `login_log` (
+    `id` MEDIUMINT UNSIGNED PRIMARY KEY AUTO_INCREMENT,
+    `login_id` MEDIUMINT UNSIGNED NOT NULL REFERENCES `login` (`id`),
+    `ip` VARCHAR(15) NOT NULL,
+    `attempts` MEDIUMINT UNSIGNED,
```
norraxx: `NOT NULL DEFAULT 1`

PR #227, opened 2020-06-05
Feature/#368 enhance avatar urls

ahsanbagwan:
Raising this PR meanwhile. V94 migration version.

### DIFF
--- a/migrations/V96__avatar-enhancements.sql
+++ b/migrations/V96__avatar-enhancements.sql
@@ -1,13 +1,13 @@
-ALTER TABLE avatars_list ADD COLUMN `filename` VARCHAR(255);
+ALTER TABLE avatars_list ADD COLUMN `filename` VARCHAR(212);
 
 UPDATE avatars_list SET filename=REPLACE(url, 'https://content.faforever.com/faf/avatars/', '');
 
-ALTER TABLE avatars_list MODIFY `filename` VARCHAR(255) NOT NULL;
+ALTER TABLE avatars_list MODIFY `filename` VARCHAR(212) NOT NULL;
 
 ALTER TABLE avatars_list DROP INDEX `url`;
 
 ALTER TABLE avatars_list DROP COLUMN `url`;
 
-ALTER TABLE `avatars_list` ADD COLUMN `url` VARCHAR(255) GENERATED ALWAYS AS (CONCAT('https://content.faforever.com/faf/avatars/', filename)) STORED;
+ALTER TABLE `avatars_list` ADD COLUMN `url` VARCHAR(255) GENERATED ALWAYS AS (CONCAT('https://content.faforever.com/faf/avatars/', filename)) VIRTUAL;
 
-ALTER TABLE `avatars_list` ADD CONSTRAINT `url` UNIQUE (`url`, `tooltip`);
+ALTER TABLE `avatars_list` ADD CONSTRAINT `filename` UNIQUE (`filename`, `tooltip`);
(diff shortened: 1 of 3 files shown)


Brutus5000:
We missed out a constraint on the table:
`constraint url unique (url, tooltip)`

This prevents deleting the url column. Needs to be dropped and recreated

migrations/V96__avatar-enhancements.sql
```diff
@@ -0,0 +1,13 @@
...
+
+UPDATE avatars_list SET filename=REPLACE(url, 'https://content.faforever.com/faf/avatars/', '');
+
+ALTER TABLE avatars_list MODIFY `filename` VARCHAR(255) NOT NULL;
+
+ALTER TABLE avatars_list DROP INDEX `url`;
+
+ALTER TABLE avatars_list DROP COLUMN `url`;
+
+ALTER TABLE `avatars_list` ADD COLUMN `url` VARCHAR(255) GENERATED ALWAYS AS (CONCAT('https://content.faforever.com/faf/avatars/', filename)) STORED;
```
norraxx: i don't think, that this column should be `stored`, it can be  (`VIRTUAL`) computed very easy and fast. And `UNIQUE` as index should be the `filename`.

migrations/V96__avatar-enhancements.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE avatars_list ADD COLUMN `filename` VARCHAR(255);
+
+UPDATE avatars_list SET filename=REPLACE(url, 'https://content.faforever.com/faf/avatars/', '');
+
+ALTER TABLE avatars_list MODIFY `filename` VARCHAR(255) NOT NULL;
+
+ALTER TABLE avatars_list DROP INDEX `url`;
+
+ALTER TABLE avatars_list DROP COLUMN `url`;
```
norraxx, line 9: Won't mysql drop index `url` automaticaly when you'll remove chained to it column?

migrations/V96__avatar-enhancements.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE avatars_list ADD COLUMN `filename` VARCHAR(255);
+
+UPDATE avatars_list SET filename=REPLACE(url, 'https://content.faforever.com/faf/avatars/', '');
+
+ALTER TABLE avatars_list MODIFY `filename` VARCHAR(255) NOT NULL;
```
norraxx: `VARCHAR(212)` (255 minus  length of `https://content.faforever.com/faf/avatars/`)

migrations/V96__avatar-enhancements.sql
Brutus5000: The index below should be changed from UNIQUE (`url`, `tooltip`) to UNIQUE(`filename`, `tooltip`), then we have the index :)

norraxx: agree

ahsanbagwan: > i don't think, that this column should be `stored`, it can be (`VIRTUAL`) computed very easy and fast. And `UNIQUE` as index should be the `filename`.

I've not used Virtual generated column. Though sounds plausible what's the rationale behind this?

migrations/V96__avatar-enhancements.sql
ahsanbagwan: This seems like a good catch. Thanks a bunch!

migrations/V96__avatar-enhancements.sql
Brutus5000: It's not really relevant in this case, but in general it's always the tradeoff between computation time (which is instant here on string concat) vs. storage (which is minimal as we have less then 1000 avatars)

Brutus5000 (approved):

PR #228, opened 2020-06-06
Populate leaderboard_rating_journal

cleborys:
To fill `leaderboard_rating_journal` with entries from old games, we only need to determine the `leaderboard_id` column, currently either the one belonging to `global` or to `ladder_1v1`.

For newer games (since mid 2014) that is easy to do by looking at the featured mod ("faf" or "ladder_1v1"), but for earlier games the ladder mod doesn't exist and we try to guess it instead.
Here "earlier" means `game_stats.id < 1996398" or `game_stats.startTime < "2014-03-22 12:59:25"`.

On my laptop the query takes about 10 minutes, so I'm not sure how feasible it is as a migration, but maybe this gives you at least the chance to give me feedback on the SQL (much of which I had to acquire for this :sweat_smile: But now at least I know that terrible things happen if you don't include an `ON` in the join, outside of sqlalchemy.)

`game_stats.id = 1996398` is the first game using the `ladder1v1` mod.
However, the first game using that mod with a `game_player_stats` entry whose `after_mean` is not `NULL` is `4401810`, two years later. In between these two, there are about 275k valid "ladder1v1" games, none of them having the `after_mean` or `after_deviation` set in their `game_player_stats` rows.
After `4401810` we have about 400k  valid "ladder1v1" games with `after_mean` set and only about 1k where it is `NULL`.

There are only about 250 games we mark as ladder games before using the featured mod, few enough that I can scroll through them and not see any false positives, judging from the game name.

Questions
 1. Is `game_featuredMods.gamemod IN ("faf", "ladder1v1")` too restrictive for games with `global` rating?
 1. There is only a couple of hundred games passing the WHERE clause with `game_stats.id < 1996398", is that expected?
 1.  Maybe the `ORDER BY` is superfluous? It would be nice if the order of `leaderboard_rating_journal.id` still reflects the historical timeline

ToDo
 - [x] Check that `game_stats.id = 1996398` is the first game with featured mod `ladder_1v1` or something reasonable like that
 - [ ] Do some sanity checks

### DIFF
--- a/migrations/V95__populate_leaderboard_rating_journal.sql
+++ b/migrations/V95__populate_leaderboard_rating_journal.sql
@@ -12,7 +12,7 @@ INSERT INTO leaderboard_rating_journal (
         WHEN
           game_player_stats.score IN (-1, 0, 1)
           AND
-          game_stats.gameName LIKE "% Vs %" COLLATE utf8_bin
+          game_stats.gameName LIKE _utf8 "% Vs %" COLLATE utf8_bin
           AND
           (
             SELECT COUNT(*)
--- a/migrations/V95__populate_leaderboard_rating_journal.sql
+++ b/migrations/V95__populate_leaderboard_rating_journal.sql
@@ -28,8 +28,10 @@ INSERT INTO leaderboard_rating_journal (
             AND game_player_stats.color IN (1, 2)
             AND game_player_stats.score IN (-1, 0, 1)
           ) = 2
-          THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
-        ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
+        THEN
+          (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
+        ELSE
+          (SELECT id from leaderboard where leaderboard.technical_name = "global")
       END AS leaderboard_id,
       game_player_stats.mean,
       game_player_stats.deviation,
@@ -66,10 +68,11 @@ INSERT INTO leaderboard_rating_journal (
 )
 SELECT
     game_player_stats.id,
-    CASE
-      WHEN game_featuredMods.gamemod = "ladder1v1" THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
-      ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
-    END AS leaderboard_id,
+    IF(
+      game_featuredMods.gamemod = "ladder1v1",
+      (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1"),
+      (SELECT id from leaderboard where leaderboard.technical_name = "global")
+    ) AS leaderboard_id,
     game_player_stats.mean,
     game_player_stats.deviation,
     CASE
--- a/migrations/V95__populate_leaderboard_rating_journal.sql
+++ b/migrations/V95__populate_leaderboard_rating_journal.sql
@@ -9,35 +9,27 @@ INSERT INTO leaderboard_rating_journal (
   SELECT
       game_player_stats.id,
       CASE
-        WHEN game_stats.id < 1996398 THEN
-          CASE
-            WHEN 
-              game_player_stats.score IN (-1, 0, 1) 
-              AND
-              game_stats.gameName LIKE _utf8 "% Vs %" COLLATE utf8_bin
-              AND
-              (
-                SELECT COUNT(*)
-                FROM game_player_stats
-                WHERE game_player_stats.gameId = game_stats.id
-              ) = 2
-              AND
-              (
-                SELECT COUNT(*)
-                FROM game_player_stats
-                WHERE game_player_stats.gameId = game_stats.id
-                AND game_player_stats.AI = FALSE
-                AND game_player_stats.color IN (1, 2)
-                AND game_player_stats.score IN (-1, 0, 1)
-              ) = 2
-              THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
-            ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
-          END
-        ELSE
-          CASE
-            WHEN game_featuredMods.gamemod = "ladder1v1" THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
-            ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
-          END
+        WHEN
+          game_player_stats.score IN (-1, 0, 1)
+          AND
+          game_stats.gameName LIKE _utf8 "% Vs %" COLLATE utf8_bin
+          AND
+          (
+            SELECT COUNT(*)
+            FROM game_player_stats
+            WHERE game_player_stats.gameId = game_stats.id
+          ) = 2
+          AND
+          (
+            SELECT COUNT(*)
+            FROM game_player_stats
+            WHERE game_player_stats.gameId = game_stats.id
+            AND game_player_stats.AI = FALSE
+            AND game_player_stats.color IN (1, 2)
+            AND game_player_stats.score IN (-1, 0, 1)
+          ) = 2
+          THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
+        ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
       END AS leaderboard_id,
       game_player_stats.mean,
       game_player_stats.deviation,
@@ -52,6 +44,8 @@ INSERT INTO leaderboard_rating_journal (
       JOIN
         game_validity ON (game_stats.validity = game_validity.id)
     WHERE
+      game_stats.id < 1996398
+      AND
       game_player_stats.after_mean IS NOT NULL
       AND
       game_player_stats.after_deviation IS NOT NULL
@@ -60,4 +54,103 @@ INSERT INTO leaderboard_rating_journal (
       AND
       game_featuredMods.gamemod IN ("faf", "ladder1v1")
     ORDER BY game_player_stats.id
+;
+
+INSERT INTO leaderboard_rating_journal (
+  game_player_stats_id,
+  leaderboard_id,
+  rating_mean_before,
+  rating_deviation_before,
+  rating_mean_after,
+  rating_deviation_after
+)
+SELECT
+    game_player_stats.id,
+    CASE
+      WHEN game_featuredMods.gamemod = "ladder1v1" THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
+      ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
+    END AS leaderboard_id,
+    game_player_stats.mean,
+    game_player_stats.deviation,
+    CASE
+      WHEN game_player_stats.after_mean IS NOT NULL
+        THEN game_player_stats.after_mean
+      WHEN (
+        SELECT inner_game_player_stats.mean
+        FROM game_player_stats inner_game_player_stats
+        JOIN game_stats inner_game_stats
+        ON (inner_game_player_stats.gameId = inner_game_stats.id)
+        WHERE
+          inner_game_stats.id > game_stats.id
+          AND
+          inner_game_player_stats.playerId = game_player_stats.playerId
+          AND
+          inner_game_stats.gameMod = game_stats.gameMod
+        ORDER BY game_stats.id
+        LIMIT 1
+      ) IS NOT NULL
+        THEN (
+          SELECT inner_game_player_stats.mean
+          FROM game_player_stats inner_game_player_stats
+          JOIN game_stats inner_game_stats
+          ON (inner_game_player_stats.gameId = inner_game_stats.id)
+          WHERE
+            inner_game_stats.id > game_stats.id
+            AND
+            inner_game_player_stats.playerId = game_player_stats.playerId
+            AND
+            inner_game_stats.gameMod = game_stats.gameMod
+          ORDER BY game_stats.id
+          LIMIT 1
+        )
+      ELSE  game_player_stats.mean
+    END AS after_mean,
+    CASE
+      WHEN game_player_stats.after_deviation IS NOT NULL
+        THEN game_player_stats.after_deviation
+      WHEN (
+        SELECT inner_game_player_stats.deviation
+        FROM game_player_stats inner_game_player_stats
+        JOIN game_stats inner_game_stats
+        ON (inner_game_player_stats.gameId = inner_game_stats.id)
+        WHERE
+          inner_game_stats.id > game_stats.id
+          AND
+          inner_game_player_stats.playerId = game_player_stats.playerId
+          AND
+          inner_game_stats.gameMod = game_stats.gameMod
+        ORDER BY game_stats.id
+        LIMIT 1
+      ) IS NOT NULL
+        THEN (
+          SELECT inner_game_player_stats.deviation
+          FROM game_player_stats inner_game_player_stats
+          JOIN game_stats inner_game_stats
+          ON (inner_game_player_stats.gameId = inner_game_stats.id)
+          WHERE
+            inner_game_stats.id > game_stats.id
+            AND
+            inner_game_player_stats.playerId = game_player_stats.playerId
+            AND
+            inner_game_stats.gameMod = game_stats.gameMod
+          ORDER BY game_stats.id
+          LIMIT 1
+        )
+      ELSE game_player_stats.deviation
+    END AS after_deviation
+  FROM
+    game_stats
+    JOIN
+      game_player_stats ON (game_player_stats.gameId = game_stats.id)
+    JOIN
+      game_featuredMods ON (game_stats.gameMod = game_featuredMods.id)
+    JOIN
+      game_validity ON (game_stats.validity = game_validity.id)
+  WHERE
+    game_stats.id >= 1996398
+    AND
+    game_validity.message = "Valid"
+    AND
+    game_featuredMods.gamemod IN ("faf", "ladder1v1")
+  ORDER BY game_player_stats.id
 ;
--- a/migrations/V95__populate_leaderboard_rating_journal.sql
+++ b/migrations/V95__populate_leaderboard_rating_journal.sql
@@ -14,7 +14,7 @@ INSERT INTO leaderboard_rating_journal (
             WHEN 
               game_player_stats.score IN (-1, 0, 1) 
               AND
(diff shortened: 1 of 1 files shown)


migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,61 @@
...
+                AND game_player_stats.AI = FALSE
+                AND game_player_stats.color IN (1, 2)
+                AND game_player_stats.score IN (-1, 0, 1)
+              ) = 2
+              THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
+            ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
+          END
+        ELSE
+          CASE
+            WHEN game_featuredMods.gamemod = "ladder_1v1" THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
```
Askaholic: Game mod is `ladder1v1` with no underscore

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,61 @@
...
+      JOIN
+        game_player_stats ON (game_player_stats.gameId = game_stats.id)
+      JOIN
+        game_featuredMods ON (game_stats.gameMod = game_featuredMods.id)
+    WHERE
+      game_player_stats.after_mean IS NOT NULL
+      AND
+      game_player_stats.after_deviation IS NOT NULL
+      AND
+      game_stats.validity = 0
```
Askaholic: How about using the enum name `'VALID'`?

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,61 @@
...
+      AND
+      game_featuredMods.gamemod IN ("faf", "ladder_1v1")
```
Askaholic: ladder1v1

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,61 @@
...
+)
+  SELECT
+      game_player_stats.id,
+      CASE
+        WHEN game_stats.id < 1996398 THEN
+          CASE
+            WHEN 
+              game_player_stats.score IN (-1, 0, 1) 
+              AND
+              game_stats.gameName LIKE "% Vs %"
```
Askaholic: Unless you collate to `utf8_bin` this will not do a case sensitive match. The main problem is there are games called stuff like `400+ vs AI` or `Aussies vs The World`, so doing a case sensitive match should cut down on false positives.

Askaholic (changes requested):
Pretty sure I came up with that date by looking at the earliest game to have the `ladder1v1` featured mod.

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: I was trying to avoid the extra join, but I guess this one literally doesn't matter :)

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,63 @@
...
+              ) = 2
+              AND
+              (
+                SELECT COUNT(*)
+                FROM game_player_stats
+                WHERE game_player_stats.gameId = game_stats.id
+                AND game_player_stats.AI = FALSE
+                AND game_player_stats.color IN (1, 2)
+                AND game_player_stats.score IN (-1, 0, 1)
+              ) = 2
```
norraxx: There is no difference between 
```sql
AND
              (
                SELECT COUNT(*)
                FROM game_player_stats
                WHERE game_player_stats.gameId = game_stats.id
              ) = 2
```
and
```sql
              AND
              (
                SELECT COUNT(*)
                FROM game_player_stats
                WHERE game_player_stats.gameId = game_stats.id
                AND game_player_stats.AI = FALSE
                AND game_player_stats.color IN (1, 2)
                AND game_player_stats.score IN (-1, 0, 1)
              ) = 2
```

Second condition is much more restrictive, so i'd remove first one.

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,63 @@
...
+)
+  SELECT
+      game_player_stats.id,
+      CASE
+        WHEN game_stats.id < 1996398 THEN
+          CASE
+            WHEN 
+              game_player_stats.score IN (-1, 0, 1) 
+              AND
+              game_stats.gameName LIKE _utf8 "% Vs %" COLLATE utf8_bin
```
norraxx: Where dif you find `LIKE _utf8`, or what it does?

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,63 @@
...
+              (
+                SELECT COUNT(*)
+                FROM game_player_stats
+                WHERE game_player_stats.gameId = game_stats.id
+                AND game_player_stats.AI = FALSE
+                AND game_player_stats.color IN (1, 2)
+                AND game_player_stats.score IN (-1, 0, 1)
+              ) = 2
+              THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
+            ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
```
norraxx: Don't do this selects, use our known value:

```
"ladder_1v1" - 2
"global" - 1
```
Less subdependant queries you do, faster it will

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,63 @@
...
+          END
+        ELSE
+          CASE
+            WHEN game_featuredMods.gamemod = "ladder1v1" THEN (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")
+            ELSE (SELECT id from leaderboard where leaderboard.technical_name = "global")
```
norraxx: same here, use constants instead of selects

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,63 @@
...
+  game_player_stats_id, 
+  leaderboard_id, 
+  rating_mean_before, 
+  rating_deviation_before, 
+  rating_mean_after, 
+  rating_deviation_after
+)
+  SELECT
+      game_player_stats.id,
+      CASE
```
norraxx, line 11: Imho, instead of big `CASE ... WHEN ... THEN ... END` you can use `IF` function, it will look like `SELECT IF(condition, value_if_true, value_if_false)`

Better readibility

norraxx:
```sql
EXPLAIN
INSERT INTO leaderboard_rating_journal (
  game_player_stats_id, 
  leaderboard_id, 
  rating_mean_before, 
  rating_deviation_before, 
  rating_mean_after, 
  rating_deviation_after
)
  SELECT
      game_player_stats.id,
      IF (
        game_stats.id < 1996398,
        IF (
          game_player_stats.score IN (-1, 0, 1) 
          AND
          game_stats.gameName LIKE _utf8 "% Vs %" COLLATE utf8_bin
          AND
          (
            SELECT COUNT(*)
            FROM game_player_stats
            WHERE game_player_stats.gameId = game_stats.id
            AND game_player_stats.AI = FALSE
            AND game_player_stats.color IN (1, 2)
            AND game_player_stats.score IN (-1, 0, 1)
          ) = 2,
          2,
          1
        ),
        IF (game_featuredMods.gamemod = "ladder1v1", 2, 1)
      ) AS leaderboard_id,
      game_player_stats.mean,
      game_player_stats.deviation,
      game_player_stats.after_mean,
      game_player_stats.after_deviation
    FROM
      game_stats
      JOIN
        game_player_stats ON (game_player_stats.gameId = game_stats.id)
      JOIN
        game_featuredMods ON (game_stats.gameMod = game_featuredMods.id)
      JOIN
        game_validity ON (game_stats.validity = game_validity.id)
    WHERE
      game_player_stats.after_mean IS NOT NULL
      AND
      game_player_stats.after_deviation IS NOT NULL
      AND
      game_validity.message = "Valid"
      AND
      game_featuredMods.gamemod IN ("faf", "ladder1v1")
    ORDER BY game_player_stats.id;
```

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: Well, the idea is that you should have two players in the game and *both* players satisfy the strict requirements. Otherwise you could get a four player game where only two players satisfy the strict checks, so that you pass nonetheless.
You could also think of this as `number_of_players_that_look_as_if_this_is_ladder == number_of_players == 2`

This is all stolen from Askaholic's query in https://github.com/FAForever/server/pull/601, by the way.

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: I was of the impression that to do the `COLLATE` I would need to enter a utf8 string literal and that this is the way to do it. That doesn't seem true so I think it should be removed :)

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: That is true. There is no easy way to save these in local variables/constants and then use these, right? Because I would prefer not to hardcode constants and rather have a
```
@ladder_leaderboard_id = (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1")

// ...

CASE ...stuff... THEN @ladder_leaderboard_id
```

But if I'm not mistaken MySQL only allows that if I wrap it in a procedure?

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: Ah that's nice (and I didn't know about it) :+1:

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: On further inspection it seems necessary since otherwise the charset is `latin1`?
https://dev.mysql.com/doc/refman/5.7/en/charset-introducer.html

(`_utf8` is the MySQL v5.7 variant instead of `_utf8mb4` for MySQL v8.0)

migrations/V95__populate_leaderboard_rating_journal.sql
norraxx: You know this values already, they are constants. It's one and two :-)

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,159 @@
...
+            inner_game_stats.id > game_stats.id
+            AND
+            inner_game_player_stats.playerId = game_player_stats.playerId
+            AND
+            inner_game_stats.gameMod = game_stats.gameMod
+          ORDER BY game_stats.id
+          LIMIT 1
+        )
+      ELSE game_player_stats.deviation
+    END AS after_deviation
```
norraxx, line 143: Oh my holy zombie jesus :-)

migrations/V95__populate_leaderboard_rating_journal.sql
norraxx: IF they satisfy second condition, they'll also satisfy the first one.

migrations/V95__populate_leaderboard_rating_journal.sql
norraxx: `The introducer is just a signal` - so i don't get why would you need it?
Do we want to have case sensitive like? If yes, maybe we should use `utf8mb4_bin`, because we have `utf8mb4_unicode_ci` in `gameName`?

Also `utf8` vs `utf8mb4` is not the same, they're not compatibile, there is a difference and support between mysql versions.

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,159 @@
...
+  rating_deviation_before,
+  rating_mean_after,
+  rating_deviation_after
+)
+SELECT
+    game_player_stats.id,
+    IF(
+      game_featuredMods.gamemod = "ladder1v1",
+      (SELECT id from leaderboard where leaderboard.technical_name = "ladder_1v1"),
+      (SELECT id from leaderboard where leaderboard.technical_name = "global")
```
norraxx, line 74: If you want to have QUERY in select part of the QUERY, how about to use `SELECT 1`? :see_no_evil:

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,159 @@
...
+          WHERE
+            inner_game_stats.id > game_stats.id
+            AND
+            inner_game_player_stats.playerId = game_player_stats.playerId
+            AND
+            inner_game_stats.gameMod = game_stats.gameMod
+          ORDER BY game_stats.id
+          LIMIT 1
+        )
+      ELSE game_player_stats.deviation
```
norraxx, line 142: First of all, this is typical case for using `coalesce`:

`COALESCE(game_player_stats.after_deviation, inner_game_player_stats.deviation, game_player_stats.deviation)`

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,159 @@
...
+          ON (inner_game_player_stats.gameId = inner_game_stats.id)
+          WHERE
+            inner_game_stats.id > game_stats.id
+            AND
+            inner_game_player_stats.playerId = game_player_stats.playerId
+            AND
+            inner_game_stats.gameMod = game_stats.gameMod
+          ORDER BY game_stats.id
+          LIMIT 1
+        )
```
norraxx, line 141: I don't undrstand, why would you need that double query? Like getting next deviation that is bigger than this current result, for current player and mod. O_o. What are you trying to solve?

And If you need it so badly, why would you do it as dependant query, why not to join it bellow as 2x left join (i'll write)

migrations/V95__populate_leaderboard_rating_journal.sql
```diff
@@ -0,0 +1,159 @@
...
+      ELSE game_player_stats.deviation
+    END AS after_deviation
+  FROM
+    game_stats
+    JOIN
+      game_player_stats ON (game_player_stats.gameId = game_stats.id)
+    JOIN
+      game_featuredMods ON (game_stats.gameMod = game_featuredMods.id)
+    JOIN
+      game_validity ON (game_stats.validity = game_validity.id)
```
norraxx, line 151: continue from above
```sql
LEFT JOIN game_player_stats inner_game_player_stats 
ON (
	inner_game_player_stats.playerId = game_player_stats.playerId
)
LEFT JOIN game_stats inner_game_stats 
ON (
		inner_game_player_stats.gameId = inner_game_stats.id
	AND inner_game_stats.id > game_stats.id
	AND inner_game_stats.gameMod = game_stats.gameMod
)

```

norraxx:
When i see it, maybe we should process this results like in bulks with python script? It won't be blocking operation, in small transactions, just go one by one... Commit like every 1000 rows...

migrations/V95__populate_leaderboard_rating_journal.sql
Askaholic: Not true. Consider a game with 2 human players which satisfy all conditions and one AI player. The second query will return a count of 2 but the first query will return a count of 3.

migrations/V95__populate_leaderboard_rating_journal.sql
Askaholic: Isn't ASCII the same in latin1 and utf8? Since the literal is just ASCII I don't think it matters.

migrations/V95__populate_leaderboard_rating_journal.sql
Askaholic: Strings are better. They are more readable and the table is tiny. There will be no performance cost.

migrations/V95__populate_leaderboard_rating_journal.sql
Askaholic: See discussion on zulip in `#tmm-discussions`

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: So what do you suggest? Simply removing `_utf8` makes the query fail with `COLLATION 'utf8_bin' is not valid for CHARACTER SET 'latin1'`, see https://travis-ci.org/github/FAForever/db/builds/695789080.

And I thought `utf8mb4` is not yet supported in MySQL v3.7 which the docker image is using?

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: One sentence summary: There are two years worth of of `game_player_stats` rows for ladder games, which have nulled out the `after_mean`, so we try to look it up from the next game the player played instead.

migrations/V95__populate_leaderboard_rating_journal.sql
norraxx: There is performance cost. Each dependant subquery is bad.
We should get data in join phase.

Brutus5000:
From the supcomhub migration stuff I have this leftover repair migration:

```sql
UPDATE faf.game_stats SET gamemod = 6
WHERE gamemod = 0 AND
        startTime < '2014-03-22 12:59:25' AND
        gameName LIKE '% Vs %' AND
        (
            SELECT count(*) FROM faf.game_player_stats
            WHERE gameId = faf.game_stats.id
        ) = 2
  AND
        (
            SELECT count(*) FROM faf.game_player_stats
            WHERE gameId = faf.game_stats.id AND
                    AI = false AND
                    color IN (1, 2) AND
                    score in (-1, 0, 1)
        ) = 2;
```

Running this first would make things a lot easier and the other queries less complex.

cleborys:
Seeing that there are two years of games which don't have the `after_mean` filled in, as well as a couple of rating hiccups, where games were rated out of order and overwrote previous rating changes, after talking to Askaholic I'd suggest that instead of trying to copy data over we re-rate the entire game history.
Once we have an independent rating service that should be a lot easier by simply "replaying" the appropriate messages for all valid past games.

Askaholic:
Oh yea we should definitely run that sort of repair migration first. We probably should have done that when I tried to fix the win ratio before, but oh well.

cleborys:
Closed in favour of re-rating the entire history in a batch job, necessary cleanup moved to #231.

migrations/V95__populate_leaderboard_rating_journal.sql
cleborys: Oh that's very nice.

PR #230, opened 2020-06-09
Issue/#229 Add foreign key constraints

### LINKED ISSUE #229: Fix foreign key constraints
Turns out mysql doesn't honor inline `REFERENCES` declarations, so we need to go back and add foreign key constraints manually for these tables:

- `leaderboard_rating`
- `leaderboard_rating_journal`
- `matchmaker_queue`
- `map_pool_map_version`
- `matchmaker_queue_map_pool`
- `map_reviews_summary`
- `mod_reviews_summary`

I think this has been a pretty common problem, and the cause of several issues over the years as there are a lot of migrations already for adding in missing foreign key constraints.

Askaholic:
I think that any existing data will not be violating the constraints (tables should be mostly empty though). Not sure about `map_reviews_summary` and `mod_reviews_summary` though.

Fixes #229

### DIFF
--- a/migrations/v95__add_foreign_key_constraints.sql
+++ b/migrations/v95__add_foreign_key_constraints.sql
@@ -1,0 +1,25 @@
+ALTER TABLE leaderboard_rating
+  ADD FOREIGN KEY (login_id) references login (id)
+  ADD FOREIGN KEY (leaderboard_id) references leaderboard_id (id);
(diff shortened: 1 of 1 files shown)


norraxx (commented):
LGTM

Brutus5000 (approved):

PR #231, opened 2020-06-09
Old game cleanup

cleborys:
The first query sets the featured mod of old ladder games where the `ladder1v1` mod didn't yet exist, but where we can be reasonably confident they were ladder. (Discussed in many places, most recently in https://github.com/FAForever/db/pull/228.)

The second query finds games where the associated `game_player_stats` rows don't have the `playerId` filled in correctly with the id of the same player repeated every row.
Here is an example: 
```
mysql> select * from game_player_stats where gameId = 4393013;                                            
+---------+---------+----------+----+---------+-------+------+-------+---------+-----------+------------+-----------------+-------+---------------------+---------+
| id      | gameId  | playerId | AI | faction | color | team | place | mean    | deviation | after_mean | after_deviation | score | scoreTime           | result  |
+---------+---------+----------+----+---------+-------+------+-------+---------+-----------+------------+-----------------+-------+---------------------+---------+
| 8183192 | 4393013 |   150494 |  0 |       1 |     3 |    3 |     7 | 676.332 |   207.979 |       NULL |            NULL |    -1 | 2016-02-03 15:47:13 | UNKNOWN |
| 8183193 | 4393013 |   150494 |  0 |       4 |     5 |    2 |     1 | 1376.66 |   195.131 |       NULL |            NULL |    -1 | 2016-02-03 15:47:13 | UNKNOWN |
+---------+---------+----------+----+---------+-------+------+-------+---------+-----------+------------+-----------------+-------+---------------------+---------+

 mysql> select * from game_stats where id = 4393013;
+---------+---------------------+---------------------+--------------+----------+---------+--------+-------+------------------------+----------+
| id      | startTime           | endTime             | replay_ticks | gameType | gameMod | host   | mapId | gameName               | validity |
+---------+---------------------+---------------------+--------------+----------+---------+--------+-------+------------------------+----------+
| 4393013 | 2016-02-03 15:47:13 | 2016-02-03 15:47:18 |         NULL | 3        |       0 | 133570 |  NULL |   [PLAYER NAME]'s game |        2 |
+---------+---------------------+---------------------+--------------+----------+---------+--------+-------+------------------------+----------+
```
This seems to have been a short-lived bug for a few weeks around February 2016, around the time of the introduction of writing rating results in the `after_mean` column for the `ladder1v1 mod` (e.g. the first `ladder1v1` game which has that column filled in is affected).
Over all, about 20k valid `game_player_stats` rows are affected and to prevent these from messing with re-calculating rating history I would suggest we set their validity to `"Unranked for another reason"`.
To speed up the query I restricted it to the range of (valid) affected games:
```
mysql> select min(game_stats.id) AS min_game_id, max(game_stats.id) AS max_game_id , count(*) AS game_count FROM game_stats LEFT JOIN game_player_stats ON (game_stats.id = game_player_stats.gameId) WHERE game_stats.validity = 0 AND (SELECT count(*) from game_player_stats where game_player_stats.gameId = game_stats.id) != (SELECT count(DISTINCT playerId) from game_player_stats where game_player_stats.gameId = game_stats.id);
+-------------+-------------+------------+
| min_game_id | max_game_id | game_count |
+-------------+-------------+------------+
|     4393128 |     4424726 |      18219 |
+-------------+-------------+------------+
1 row in set (3 min 32.63 sec)
```
(The example game above is notably outside of that range, but the validity is already set to invalid (not assassination mode).)


migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,44 @@
+UPDATE
+  faf.game_stats
+SET
+  gamemod = 6
+WHERE
+  gamemod = 0
+AND
+  startTime < "2014-03-22 12:59:25"
+AND
+  gameName LIKE _utf8 "% Vs %" COLLATE utf8_bin
```
cleborys: I still don't understand how to do case sensitive matching without the `_utf8`, feel free to let me know what I'm missing :kissing_heart:

Brutus5000: The collation of the column is case insensitive, so you shouldn't need to?

migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,44 @@
...
+    SELECT count(*) FROM faf.game_player_stats
+    WHERE gameId = faf.game_stats.id
+  ) != (
+    SELECT count(DISTINCT faf.game_player_stats.playerId) FROM faf.game_player_stats
+    WHERE gameId = faf.game_stats.id
+  )
+AND
+  faf.game_stats.id > 4393127
+AND
+  faf.game_stats.id < 4424727
```
Brutus5000: There is this magic operator called between.

migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,44 @@
...
+  validity = 0
+AND
+  (
+    SELECT count(*) FROM faf.game_player_stats
+    WHERE gameId = faf.game_stats.id
+  ) != (
+    SELECT count(DISTINCT faf.game_player_stats.playerId) FROM faf.game_player_stats
+    WHERE gameId = faf.game_stats.id
+  )
+AND
```
Brutus5000: From my experience these kind of subselects run forever. Can you put the the subselect results into a temporary table and join them=

migrations/V95__laddermod_cleanup.sql
cleborys: magic indeed.

migrations/V95__laddermod_cleanup.sql
norraxx: He need to have a exact match, to know that's game name is not made by user. (it's not 100% sure)

On test i see:
```sql
`gameName` varchar(128) COLLATE utf8mb4_unicode_ci NOT NULL
```
So i would use  `utf8mb4_bin`, not `utf8_bin`, reason is number of bytes per character, it can be different.
And i still don't get this `_utf8`, so you can remove it with `utf8mb4_bin`

migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,44 @@
...
+  faf.game_stats.id > 4393127
+AND
+  faf.game_stats.id < 4424727
+;
```
norraxx: Please, remove all database prefixes (`faf.`). This query won't work on test.

migrations/V95__laddermod_cleanup.sql
cleborys: I'm learning something new with every query I write! Let's see if I understood you correctly :P

norraxx (changes requested):

migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,54 @@
...
+  faf.game_stats
+INNER JOIN
+  faf.game_player_count ON (faf.game_stats.id = faf.game_player_count.id)
+SET
+  validity = 24  # "Unranked for another reason"
+WHERE
+  validity = 0
+AND
+  gps_row_count != distinct_player_count
+;
```
norraxx: `DROP TEMPORARY TABLE faf.game_player_count;` :-)

migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,54 @@
...
+UPDATE
+  faf.game_stats
+INNER JOIN
+  faf.game_player_count ON (faf.game_stats.id = faf.game_player_count.id)
+SET
+  validity = 24  # "Unranked for another reason"
+WHERE
+  validity = 0
+AND
+  gps_row_count != distinct_player_count
```
norraxx: it's good to have from which table did you get this column. Like mysql will resolve it, but my eyes not :-D

migrations/V95__laddermod_cleanup.sql
```diff
@@ -0,0 +1,54 @@
...
+      count(faf.game_player_stats.id) AS gps_row_count,
+      count(DISTINCT faf.game_player_stats.playerId) AS distinct_player_count
+    FROM 
+      faf.game_stats
+    INNER JOIN 
+      faf.game_player_stats ON (faf.game_player_stats.gameId = faf.game_stats.id)
+    WHERE
+      faf.game_stats.id BETWEEN 4393127 AND 4424727
+    GROUP BY
+      faf.game_stats.id
```
norraxx: Hmm, i think, you could filter results by clause `HAVING`, which is applied after `group by`, so you'll have your table smaller.

`HAVING gps_row_count != distinct_player_count`

norraxx: And ofc you won't need that condition in query bellow

norraxx:
@cleborys @Askaholic Guys, you'll have conflicts with migration in names #230 #231. Also `V96__` is used by #227.

migrations/V95__laddermod_cleanup.sql
cleborys: Here's an example where case sensitivity matters
```
mysql> select game_stats.id, game_stats.gameName from game_stats where gameName LIKE "% Vs %" and gameName not LIKE _utf8 "% Vs %" COLLATE utf8_bin;
+---------+------------+
| id      | gameName   |
+---------+------------+
| // ...               |
| 6169271 | ZLO 4 vs 1 |
| // ...               |
+---------+------------+
```

cleborys: What I didn't understand before is that you don't need to make the literal string `_utf8` if you collate the column instead of the string :+1:

cleborys:
I renamed it to `V97__` so the other two PRs can use `V95__` and `V96__`.

Askaholic:
Tbh renaming files is trivial. My approach is to always use the next available number and rebase my PR if something else is merged. Yea it can delay by 1 day or something but that’s not a big deal here at FAF

migrations/V97__laddermod_cleanup.sql
```diff
@@ -0,0 +1,56 @@
+UPDATE
+  game_stats
+SET
+  gamemod = 6
```
norraxx: `gameMod`

migrations/V97__laddermod_cleanup.sql
```diff
@@ -0,0 +1,56 @@
...
+WHERE
+  gamemod = 0
+AND
+  startTime < "2014-03-22 12:59:25"
+AND
+  gameName COLLATE utf8mb4_bin LIKE "% Vs %"
+AND
+  (
+    SELECT count(*) FROM game_player_stats
+    WHERE gameId = game_stats.id
```
norraxx: [proposition for better code]

I would preffer to read code like

```
select part
tables part + join filters
conditions part
order part
having part.
```
It's easier to read when it's on different lines.

\+ keywords like function names, etc.. could be ca(ra)melized. :-)
[/proposition for better code]

migrations/V97__laddermod_cleanup.sql
```diff
@@ -0,0 +1,56 @@
...
+  ) = 2
```
norraxx: I still don't think, that you need that first condition. Sorry, but your second query to the table is more restrictive. And mostly it will do whole job that you'll need. Query above will be still filtered by query bellow.

Let's have a talk some time about it?

norraxx (commented):
I suppose your code will work, unfortunately i don't have a data on local machine. I'm offering to sit some evening together and make some queries together.

migrations/V97__laddermod_cleanup.sql
Askaholic: You need both conditions:

Consider a game with 2 human players which satisfy all conditions and one AI player. The second query will return a count of 2 but the first query will return a count of 3.

Brutus5000:
@cleborys Is this work in progress or do you consider this finished? In the latter case, please move to the version 103

cleborys:
I have added noraxx' formatting suggestions and rebased, I believe this is ready.

Brutus5000 (approved):

PR #234, opened 2020-06-23
Ranks for leaderboard

### LINKED ISSUE #233: Make materialized view with ranks for leaderboards

1-alex98:
Fixes #233

### DIFF
--- a/migrations/V97__rating_materialized_view.sql
+++ b/migrations/V97__rating_materialized_view.sql
@@ -26,17 +26,17 @@ ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
 COMMENT 'Fill the materialized view including ranks for ladder'
 DO
 BEGIN
-    CREATE TEMPORARY TABLE active_players ENGINE=MEMORY AS
+    CREATE TEMPORARY TABLE active_ladder_players ENGINE=MEMORY AS
     (
         SELECT DISTINCT gps.playerId
         FROM game_player_stats gps
-                 INNER JOIN game_stats gs on gps.gameId = gs.id
+        INNER JOIN game_stats gs on gps.gameId = gs.id
         WHERE gs.endTime > now() - INTERVAL 1 YEAR
           AND gs.gameMod = 6 -- ladder
           AND gs.validity = 0
     );
 
-    DELETE FROM active_players
+    DELETE FROM active_ladder_players
     WHERE playerId IN (
         SELECT player_id from ban
         WHERE (expires_at is null or expires_at > NOW()) AND revoke_time IS NULL
@@ -53,10 +53,10 @@ BEGIN
            winGames,
            rating
     FROM ladder1v1_rating
-    INNER JOIN active_players ON ladder1v1_rating.id = active_players.playerId
+    INNER JOIN active_ladder_players ON ladder1v1_rating.id = active_ladder_players.playerId
     ORDER BY rating DESC;
 
-    DROP TABLE active_players;
+    DROP TABLE active_ladder_players;
 
     COMMIT;
 END $$
@@ -66,17 +66,17 @@ ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
 COMMENT 'Fill the materialized view including ranks for global'
 DO
 BEGIN
-    CREATE TEMPORARY TABLE active_players ENGINE=MEMORY AS
+    CREATE TEMPORARY TABLE active_global_players ENGINE=MEMORY AS
     (
         SELECT DISTINCT gps.playerId
         FROM game_player_stats gps
-                 INNER JOIN game_stats gs on gps.gameId = gs.id
+        INNER JOIN game_stats gs on gps.gameId = gs.id
         WHERE gs.endTime > now() - INTERVAL 1 YEAR
           AND gs.gameMod = 0 -- global
           AND gs.validity = 0
     );
 
-    DELETE FROM active_players
+    DELETE FROM active_global_players
     WHERE playerId IN (
         SELECT player_id from ban
         WHERE (expires_at is null or expires_at > NOW()) AND revoke_time IS NULL
@@ -92,12 +92,10 @@ BEGIN
            numGames,
            rating
     FROM global_rating
-    LEFT JOIN ban ON global_rating.id = ban.player_id
-    WHERE ((ban.expires_at is null or ban.expires_at > NOW()) AND ban.revoke_time IS NULL)
-      AND  is_active = 1
+    INNER JOIN active_global_players ON global_rating.id = active_global_players.playerId
     ORDER BY rating DESC;
 
-    DROP TABLE active_players;
+    DROP TABLE active_global_players;
 
     COMMIT;
 END $$
--- a/migrations/V97__rating_materialized_view.sql
+++ b/migrations/V97__rating_materialized_view.sql
@@ -1,0 +1,75 @@
+CREATE TABLE IF NOT EXISTS ladder1v1_rating_rank_view
+(
+    id        mediumint unsigned             not null,
+    ranking   int,
+    mean      float                          null,
+    deviation float                          null,
+    numGames  smallint(4) unsigned default 0 not null,
+    winGames  smallint(4) unsigned default 0 not null,
+    is_active tinyint(1)           default 0 not null,
+    rating    float
+);
+
+CREATE TABLE IF NOT EXISTS global_rating_rank_view
+(
+    id        mediumint unsigned             not null,
+    ranking   int,
+    mean      float                          null,
+    deviation float                          null,
+    numGames  smallint(4) unsigned default 0 not null,
+    winGames  smallint(4) unsigned default 0 not null,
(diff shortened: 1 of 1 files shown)


migrations/V97__rating_materialized_view.sql
```diff
@@ -0,0 +1,75 @@
...
+DELIMITER $$
+
+CREATE EVENT IF NOT EXISTS `global_calculate_ranks_view`
+    ON SCHEDULE EVERY 5 MINUTE STARTS CURRENT_TIMESTAMP
+    COMMENT 'Fill the materialized view including ranks for global'
+    DO
+    BEGIN
+        START TRANSACTION;
+        SET @r = 0;
+        DELETE FROM ladder1v1_rating_rank_view;
```
bukajsytlos: global

bukajsytlos (changes requested):

bukajsytlos (changes requested):
Original "view" ignored banned players

migrations/V97__rating_materialized_view.sql
```diff
@@ -0,0 +1,75 @@
+create table if not exists ladder1v1_rating_rank_view
+(
+    id        mediumint unsigned             not null,#
```
norraxx: `#` ?

migrations/V97__rating_materialized_view.sql
```diff
@@ -0,0 +1,75 @@
...
+    ON SCHEDULE EVERY 5 MINUTE STARTS CURRENT_TIMESTAMP
+    COMMENT 'Fill the materialized view including ranks for ladder'
+    DO
+    BEGIN
+        START TRANSACTION;
+        SET @r = 0;
+        DELETE FROM ladder1v1_rating_rank_view;
+        INSERT INTO ladder1v1_rating_rank_view(id, ranking, mean, deviation, numGames, winGames, is_active, rating)
+        SELECT id,
+               @r := (@r + 1),
```
norraxx: What's the difference between some auto incremented column and this magic that replaces function from mysql 8 `row_number()`. You flush table always, so add there some auto incremented column, set it to 0 and do anything you want without session values, that can be common for both events.

norraxx (changes requested):

norraxx:
> Tell me how to improve

Use big letters, it's nice to read after that

norraxx:
> I copied from stack overflow so I guess I should use a local variable

See auto incremented column solution above :joy:

norraxx:
> I copied from stack overflow so I guess I should use a local variable

Yes, i understand and for session and simple querying to database it will work well. Unfortunately i don't know how session variables behave during events and if two events are running same time, and have same named variable, how they will behave.

migrations/V97__rating_materialized_view.sql
1-alex98: I think we dont have mysql 8

1-alex98: There is also a RANK() function but well we are not mysql 8 it seems

1-alex98: Yeah nice if we could do it together one day

norraxx: Sure! Let's do it today or tomorrow evening.

1-alex98: today sounds good

migrations/V97__rating_materialized_view.sql
1-alex98: yes

migrations/V97__rating_materialized_view.sql
1-alex98: yes

migrations/V97__rating_materialized_view.sql
1-alex98: Well I did research and i can not use these local variables. They don't offer setting them inside a query. So I need to stick with session variables. But I called them differently so it avoids any collisions.

norraxx (commented):
I'm sexy and I know it.

norraxx (approved):

1-alex98:
https://github.com/FAForever/faf-java-api/pull/405

migrations/V97__rating_materialized_view.sql
```diff
@@ -0,0 +1,67 @@
...
+BEGIN
+    ALTER TABLE global_rating_rank_view AUTO_INCREMENT = 1, ALGORITHM=INPLACE;
+    START TRANSACTION;
+    DELETE FROM global_rating_rank_view;
+    INSERT INTO global_rating_rank_view(`id`, `mean`, `deviation`, `num_games`, `win_games`, `rating`)
+    SELECT id,
+           mean,
+           deviation,
+           numGames,
+           0,
```
Brutus5000: Global rating doesn't have won games. Putting in 0 value isn't making it better.

Brutus5000 (changes requested):

migrations/V97__rating_materialized_view.sql
norraxx: Do you propose null value? Like it what got into our heads with alex.

Brutus5000: I propose to drop the column so nobody queries wrong stuff in the API.

norraxx: @axel1200 ?

1-alex98: ok, just thought I leave it there because should be there. Or maybe put null.

1-alex98:
Removed it

migrations/V97__rating_materialized_view.sql
```diff
@@ -26,19 +26,38 @@ ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
 COMMENT 'Fill the materialized view including ranks for ladder'
 DO
 BEGIN
+    CREATE TEMPORARY TABLE active_players ENGINE=MEMORY AS
+    (
+        SELECT DISTINCT gps.playerId
+        FROM game_player_stats gps
+                 INNER JOIN game_stats gs on gps.gameId = gs.id
```
norraxx: formatting :-)

migrations/V97__rating_materialized_view.sql
```diff
@@ -47,18 +66,39 @@ ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
...
+    SELECT global_rating.id,
            mean,
            deviation,
            numGames,
            rating
     FROM global_rating
-    WHERE is_active = 1
+    LEFT JOIN ban ON global_rating.id = ban.player_id
+    WHERE ((ban.expires_at is null or ban.expires_at > NOW()) AND ban.revoke_time IS NULL)
+      AND  is_active = 1
```
norraxx: :thinking: You don't use `active_players` in this event.

Also I don't understand this condition.
`((ban.expires_at is null or ban.expires_at > NOW()) AND ban.revoke_time IS NULL)`
Infinite ban expire, or ban is in future and ban is not revoked.
You want to join and have only banned players? O_o

Imho copy-paste kung-fu :see_no_evil:

migrations/V97__rating_materialized_view.sql
```diff
@@ -47,18 +66,39 @@ ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
 COMMENT 'Fill the materialized view including ranks for global'
 DO
 BEGIN
+    CREATE TEMPORARY TABLE active_players ENGINE=MEMORY AS
```
norraxx: Also when i see this, I'm thinking about to have an materialized view for active players, refreshed once per day.
I've seen this query with such conditions already on some places.

migrations/V97__rating_materialized_view.sql
```diff
@@ -47,18 +66,39 @@ ON SCHEDULE EVERY 1 HOUR STARTS CURRENT_TIMESTAMP
...
+    (
+        SELECT DISTINCT gps.playerId
+        FROM game_player_stats gps
+                 INNER JOIN game_stats gs on gps.gameId = gs.id
```
norraxx: formatting

migrations/V97__rating_materialized_view.sql
norraxx: Also you're making table 2 times with same name `active_players` at two places, but data are differrent. I'd name them different, to avoid undefined behavior between two events run.

norraxx (changes requested):

migrations/V97__rating_materialized_view.sql
Brutus5000: Yes, stupid. Fixed now.

migrations/V97__rating_materialized_view.sql
Brutus5000: I only ever saw querying active players for the leaderboard.

Brutus5000: I renamed the 2 temp tables now to avoid collisions as suggested.

norraxx (approved):

1-alex98:
Approve as well

Brutus5000 (approved):

Issue #236, opened 2020-07-06
Change V97 tables/materialized views gather the data from leaderboard_rating


norraxx:
Problem:
we've made unwanted migration from `ladder1v1_rating` and `global_rating` to point data to new table `leaderboard_rating`.

Goal:
create views `ladder1v1_rating_rank_view` and `global_rating_rank_view` that will gather data from new table `leaderboard_rating`

Brutus5000:
@norraxx you need to add "closes #236" or "fixes #236" in the commit message to trigger auto-closing the issue after merge.

PR #238, opened 2020-07-06
Migrate to new leaderboard rating (fixes #237)

Brutus5000:

PR #239, opened 2020-07-06
#236 using different table for leaderboard

norraxx:


Brutus5000 (approved):

norraxx:
@Brutus5000 Changes on test worked. Still this migration should be applied after #238, otherwise see #238

PR #242, opened 2020-07-18
Add id column to table map_pool_map_version

Brutus5000:


migrations/V100__map_pool_map_version_add_id.sql
```diff
@@ -0,0 +1,21 @@
+-- From a db perspective the table map_pool_map_version doesn't need an id column.
+-- However accessing it via JSONAPI/Elide we can't use a composite primary key.
+
+CREATE TEMPORARY TABLE map_pool_map_version_archive
+SELECT * FROM map_pool_map_version;
```
norraxx, line 5: Why not to rename previous table?

Brutus5000:
As it turns out this is not required in Elide for m:n relationships. I got stuck, but the actual problem is elsewhere.

PR #243, opened 2020-07-31
Promote table matchmaker_queue_map_pool to full entity

Brutus5000:
Adds auto incrementing id and create/update time.

### DIFF
--- a/migrations/V100__promote_mmq_mappool_to_entity.sql
+++ b/migrations/V100__promote_mmq_mappool_to_entity.sql
@@ -1,0 +1,21 @@
+ALTER TABLE matchmaker_queue_map_pool
+    DROP FOREIGN KEY matchmaker_queue_map_pool_ibfk_1;
+
+ALTER TABLE matchmaker_queue_map_pool
+    DROP FOREIGN KEY matchmaker_queue_map_pool_ibfk_2;
+
+
+ALTER TABLE matchmaker_queue_map_pool DROP PRIMARY KEY;
+
+ALTER TABLE matchmaker_queue_map_pool
+    ADD id INTEGER UNSIGNED AUTO_INCREMENT FIRST,
+    ADD create_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    ADD update_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL ON UPDATE CURRENT_TIMESTAMP,
+    ADD CONSTRAINT matchmaker_queue_map_pool_pk
+        PRIMARY KEY (id),
+    ADD CONSTRAINT matchmaker_queue_map_pool_map_pool_id_fk
+        FOREIGN KEY (map_pool_id) REFERENCES map_pool (id),
+    ADD CONSTRAINT matchmaker_queue_map_pool_matchmaker_queue_id_fk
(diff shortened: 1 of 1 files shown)


migrations/V100__promote_mmq_mappool_to_entity.sql
```diff
@@ -0,0 +1,22 @@
...
+
+alter table matchmaker_queue_map_pool
+    drop foreign key matchmaker_queue_map_pool_ibfk_2;
+
+alter table matchmaker_queue_map_pool drop primary key;
+
+alter table matchmaker_queue_map_pool
+    add id int unsigned auto_increment first,
+    add create_time timestamp default CURRENT_TIMESTAMP not null,
+    add update_time timestamp default CURRENT_TIMESTAMP not null,
```
norraxx: I'm not sure about current time stamps (update & create).
Don't hurry, I'll check it, how it works. :-)

migrations/V100__promote_mmq_mappool_to_entity.sql
```diff
@@ -0,0 +1,22 @@
+-- The table matchmaker_queue_map_pool has attributes and is no simple m:n table.
+-- Therefore it needs to be promoted to a dedicated entity in order to work with JPA.
+
+alter table matchmaker_queue_map_pool
+    drop foreign key matchmaker_queue_map_pool_ibfk_1;
+
+alter table matchmaker_queue_map_pool
+    drop foreign key matchmaker_queue_map_pool_ibfk_2;
```
norraxx: Why do you drop foreign key constraints? Like, how it supposed to help us? And what it has to solve?

norraxx (commented):
How about to talk about this change? :-)

migrations/V100__promote_mmq_mappool_to_entity.sql
norraxx: try something like that
`add update_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

More info in docs: https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html

migrations/V100__promote_mmq_mappool_to_entity.sql
Brutus5000: Because I can't drop the primary key otherwise.

migrations/V100__promote_mmq_mappool_to_entity.sql
Brutus5000: Yes `on update ` is missing (blame on me for using a GUI that doesn't offer this)

migrations/V100__promote_mmq_mappool_to_entity.sql
norraxx: Hmmm. Never meet such problem, that's odd.

migrations/V100__promote_mmq_mappool_to_entity.sql
norraxx: Data grid?

migrations/V100__promote_mmq_mappool_to_entity.sql
Brutus5000: Maybe your db had no data in it. It doesn't matter the table has like 5 rows inside.

norraxx (approved):
ok

PR #249, opened 2020-09-07
Add matchmaker_queue_game table

cleborys:
Adds a new m:n relationship table for linking games to their originating matchmaker queue, as suggested in https://github.com/FAForever/server/issues/599.
This is needed to properly find the recent game history per queue to avoid map repetition.
PR https://github.com/FAForever/server/pull/654 depends on this table.

### DIFF
--- a/migrations/V103__link_matchmaker_games_to_originating_queue.sql
+++ b/migrations/V103__link_matchmaker_games_to_originating_queue.sql
@@ -1,0 +1,9 @@
+CREATE TABLE matchmaker_queue_game
+(
+  id                  INT(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  matchmaker_queue_id INT(10) UNSIGNED NOT NULL,
+  game_stats_id       INT(10) UNSIGNED NOT NULL,
+  FOREIGN KEY (matchmaker_queue_id) REFERENCES matchmaker_queue (id),
+  FOREIGN KEY (game_stats_id) REFERENCES game_stats (id),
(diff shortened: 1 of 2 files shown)


migrations/V103__link_matchmaker_games_to_originating_queue.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE matchmaker_queue_game
+(
+  id                  INT(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  matchmaker_queue_id INT(10) UNSIGNED NOT NULL,
+  game_stats_id       INT(10) UNSIGNED NOT NULL,
+  FOREIGN KEY (matchmaker_queue_id) REFERENCES matchmaker_queue (id),
+  FOREIGN KEY (game_stats_id) REFERENCES game_stats (id),
+  UNIQUE INDEX (game_stats_id, matchmaker_queue_id)
```
Askaholic: If we wanted to we could enforce a one-to-many relationship by making `game_stats_id` unique. Not sure if we want that, but to me it doesn't really make sense for a single game to be somehow linked with two different matchmaker queues. Food for thought.

Brutus5000:
If it's just a m:n table without additional filter properties, we don't need an id column.

migrations/V103__link_matchmaker_games_to_originating_queue.sql
cleborys: I'm happy with either.

cleborys:
Removed the `id` column.
Note that it's now `V104`, skipping `V103`!

PR #250, opened 2020-09-24
Fix rating sign

Askaholic:
For some reason the `mean` field on `game_player_stats` is set to unsigned, even though negative mean is OK. (Even `after_mean` is signed...). Really we want to make sure that DEVIATION is unsigned. Technically speaking I think deviation must always be strictly positive, deviation of 0 will cause issues, but I'm not sure it's possible or necessary to enforce that in MySQL.

The `ladder1v1_rating` and `global_rating` tables both allow negative deviation, and so do their corresponding rank views, but since those tables are more or less deprecated anyways I just left them.

### DIFF
--- a/migrations/V105__fix_rating_signedness.sql
+++ b/migrations/V105__fix_rating_signedness.sql
@@ -1,0 +1,10 @@
+-- Mean may be negative (signed) but deviation must be positive (unsigned)
+ALTER TABLE leaderboard_rating MODIFY deviation FLOAT UNSIGNED NOT NULL;
+
+ALTER TABLE leaderboard_rating_journal
+  MODIFY rating_deviation_before FLOAT UNSIGNED NOT NULL,
+  MODIFY rating_deviation_after FLOAT UNSIGNED NOT NULL;
+
+ALTER TABLE game_player_stats
+  MODIFY mean FLOAT NOT NULL,
+  MODIFY after_deviation FLOAT UNSIGNED;


migrations/V105__fix_rating_signedness.sql
```diff
@@ -0,0 +1,10 @@
+-- Mean may be negative (signed) but deviation must be positive (unsigned)
+ALTER TABLE leaderboard_rating MODIFY deviation FLOAT UNSIGNED NOT NULL;
+
+ALTER TABLE leaderboard_rating_journal
+  MODIFY rating_deviation_before FLOAT UNSIGNED NOT NULL
+  MODIFY rating_deviation_after FLOAT UNSIGNED NOT NULL;
+
+ALTER TABLE game_player_stats
+  MODIFY mean FLOAT NOT NULL
+  MODIFY after_deviation FLOAT UNSIGNED;
```
norraxx, line 10: 1. What FLOAT units are good for? I'm just curious why it's like that.
2. Also you're facing an error like: `Out of range value for column deviation`. I'd expect some sql that fixes the data before alter.

norraxx (changes requested):

migrations/V105__fix_rating_signedness.sql
Askaholic: I don't think any of the deviations are negative, it's just a precaution. 

This is the error that we're getting in the server:
```
Traceback (most recent call last):
  File "/code/server/games/game.py", line 720, in update_game_player_stats
    await conn.execute(game_player_stats.insert().values(query_args))
  File "/code/src/aiomysql/aiomysql/sa/connection.py", line 173, in _execute
    await cursor.execute(str(compiled), post_processed_params)
  File "/code/src/aiomysql/aiomysql/cursors.py", line 239, in execute
    await self._query(query)
  File "/code/src/aiomysql/aiomysql/cursors.py", line 457, in _query
    await conn.query(q)
  File "/code/src/aiomysql/aiomysql/connection.py", line 428, in query
    await self._read_query_result(unbuffered=unbuffered)
  File "/code/src/aiomysql/aiomysql/connection.py", line 622, in _read_query_result
    await result.read()
  File "/code/src/aiomysql/aiomysql/connection.py", line 1105, in read
    first_packet = await self.connection._read_packet()
  File "/code/src/aiomysql/aiomysql/connection.py", line 593, in _read_packet
    packet.check_error()
  File "/usr/local/lib/python3.7/site-packages/pymysql/protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "/usr/local/lib/python3.7/site-packages/pymysql/err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.DataError: (1264, "Out of range value for column 'mean' at row 3")
```

Brutus5000 (approved):

migrations/V105__fix_rating_signedness.sql
Brutus5000: I verified in the database tables, there are no current/before/after deviations < 0

PR #254, opened 2021-01-11
Migrate ratings and rating changes to new leaderboard tables

### LINKED ISSUE #253: Migrate rating changes from game_player_stats to leaderboard_rating_journal
As of now `leaderboard_rating_journal` only contains games from go live since.
We need to copy all the entries deviation/mean before/after into the table.

Do not drop the columns yet (compatibility!).

Brutus5000:
Fixes #253

### DIFF
--- a/migrations/V107__migrate_to_new_leaderboards.sql
+++ b/migrations/V107__migrate_to_new_leaderboards.sql
@@ -1,0 +1,25 @@
+-- IGNORE because there are already records in the new tables we don't want to overwrite.
+
+INSERT IGNORE INTO leaderboard_rating(login_id, mean, deviation, total_games, won_games, leaderboard_id)
+SELECT id, mean, deviation, numGames, 0, 1 FROM global_rating;
+
+INSERT IGNORE INTO leaderboard_rating(login_id, mean, deviation, total_games, won_games, leaderboard_id)
+SELECT id, mean, deviation, numGames, winGames, 2 FROM ladder1v1_rating;
+
+
+INSERT IGNORE INTO leaderboard_rating_journal(game_player_stats_id, leaderboard_id, rating_mean_before,
+                                              rating_deviation_before, rating_mean_after, rating_deviation_after)
+SELECT gps.id,
+       CASE
+           WHEN gs.gameMod = 0 THEN 1
+           WHEN gs.gameMod = 6 THEN 2
+           END,
+       mean,
+       deviation,
+       after_mean,
+       after_deviation
+FROM game_player_stats gps
+         INNER JOIN game_stats gs ON gps.gameId = gs.id
+WHERE gs.gameMod IN (0, 6)
+  AND after_mean IS NOT NULL
+  AND after_deviation IS NOT NULL;


migrations/V107__migrate_to_new_leaderboards.sql
```diff
@@ -0,0 +1,25 @@
...
+INSERT IGNORE INTO leaderboard_rating(login_id, mean, deviation, total_games, won_games, leaderboard_id)
+SELECT id, mean, deviation, numGames, winGames, 2 FROM ladder1v1_rating;
+
+
+INSERT IGNORE INTO leaderboard_rating_journal(game_player_stats_id, leaderboard_id, rating_mean_before,
+                                              rating_deviation_before, rating_mean_after, rating_deviation_after)
+SELECT gps.id,
+       CASE
+           WHEN gs.gameMod = 0 THEN 1
+           WHEN gs.gameMod = 6 THEN 2
```
Askaholic, line 15: We can do this because we already had a migration that fixed the game mode for old ladder games right?

migrations/V107__migrate_to_new_leaderboards.sql
```diff
@@ -0,0 +1,25 @@
...
+           END,
+       mean,
+       deviation,
+       after_mean,
+       after_deviation
+FROM game_player_stats gps
+         INNER JOIN game_stats gs ON gps.gameId = gs.id
+WHERE gs.gameMod IN (0, 6)
+  AND after_mean IS NOT NULL
+  AND after_deviation IS NOT NULL;
```
Askaholic, line 25: This is good enough, or do we want to check that the validity is VALID?

migrations/V107__migrate_to_new_leaderboards.sql
```diff
@@ -0,0 +1,25 @@
+-- IGNORE because there are already records in the new tables we don't want to overwrite.
+
+INSERT IGNORE INTO leaderboard_rating(login_id, mean, deviation, total_games, won_games, leaderboard_id)
+SELECT id, mean, deviation, numGames, 0, 1 FROM global_rating;
```
Askaholic, line 4: Would be cool if we could fix the `won_games` too, but that's another issue.

Askaholic (approved):
I don't see any obvious flaws.

migrations/V107__migrate_to_new_leaderboards.sql
Brutus5000: Yes to our best efforts at least, see [V103](https://github.com/FAForever/db/blob/develop/migrations/V103__laddermod_cleanup.sql)

migrations/V107__migrate_to_new_leaderboards.sql
Brutus5000: It is yes. But it's really hard to figure out actual winners retroactively as the score logic seems to have changed over time.
Feel free to open a new issue for that.

migrations/V107__migrate_to_new_leaderboards.sql
Brutus5000: Hmm not sure. I checked the db. We have shitload of old games where rating changes where applied despite being invalid

PR #257, opened 2021-02-17
New flyway baseline + better data management

Brutus5000:
Put the current db schema into a single schema based on migration V107.
- This aims to fix some compatibility issues with MariaDB and also speeds up the CI pipeline.
- Some tables are now easily manageable via flyway by just editing them. The data in the `R__` prefixed file will be kept in sync with the db (unless you delete lines manually from the file)


norraxx (approved):
Good migrations squash.
Looks good to me.

PR #260, opened 2021-03-04
Support more casual game modes

Askaholic:
This would allow us to add more team setups like FFA, or cooperative/survival as well as linking mods from the vault and not just featured mods.

The mods setup will require implementation on both the client and server, but since the client already has a mod manager I think it should be fairly easy.

Also adds a column to `leaderboard` for specifying which existing leaderboard rating to use when creating a rating for the first time. https://github.com/FAForever/server/issues/724


Askaholic:
I think I want to add `teams` (maybe call it `num_teams`) but not add the mod version stuff yet.

PR #262, opened 2021-03-12
add migration for weight and map params

Sheikah45:
This is an initial pr for changing the database for changing the map_pool_map_version table to support other map params and map weights. It was generated by my mysql to make the change however there are issues with the foreign key and primary key. Currently there is no primary key as map_version_id can now be null and other_params can be null as well.

This is tied to https://github.com/FAForever/server/pull/732#issue-581528027


Brutus5000 (approved):
We tested
* Backwards-compatible with the current server
* Backwards-compatible with the current API (& related clients)

PR #272, opened 2021-04-06
Add map_play_count trigger to record map plays

### LINKED ISSUE #270: Fix map play count trigger
The trigger for map play counts is no longer working

### LINKED ISSUE #273: map play count
**Describe the bug**
the play count on the map card is not updating from the api

**Expected behavior**
map count to increase with plays

**Screenshots**
image
.

**OS**
Windows

Sheikah45:
Fixes #270
Fixes #273

### DIFF
--- a/migrations/V116__map_play_count_trigger.sql
+++ b/migrations/V116__map_play_count_trigger.sql
@@ -1,0 +1,69 @@
+-- Add times_played to map_version
+ALTER TABLE map_version
+    ADD COLUMN times_played int unsigned NOT NULL DEFAULT 0 COMMENT 'Number of times a map version has been played. Should be updated each time a row is inserted with the map_version.id in game_stats. Values start at 0' AFTER reviews;
+
+-- Add times_played to map_version
+ALTER TABLE map
+    ADD COLUMN times_played int unsigned NOT NULL DEFAULT 0 COMMENT 'Number of times a map version has been played. Should be updated each time a row is inserted with a map_version.map.id in game_stats Values start at 0' AFTER reviews;
+
+-- one time update of table_map_features to get times_played to the right play_count
+UPDATE table_map_features,
+    (SELECT mapId, COUNT(DISTINCT id) AS map_play_count
+    FROM game_stats
+    GROUP BY mapId) as map_plays
+SET table_map_features.times_played = map_plays.map_play_count
+WHERE table_map_features.map_id = map_plays.mapId;
+
+INSERT IGNORE INTO table_map_features(map_id,voters,times_played)
+SELECT mapId, '', COUNT(DISTINCT id) AS map_play_count FROM game_stats WHERE mapId IS NOT NULL GROUP BY mapId;
+
+-- one time update of map_version to get times_played to the right play_count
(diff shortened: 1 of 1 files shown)


Brutus5000:
Table_map_features should be removed (It's not used by any service) and the column should go into map_version.
Also all values should be recomputed once to be correct

Sheikah45:
Just so I am clear should that recomputation be a part of this migration

Sheikah45:
Hmm actually the api does use table_map_features it just calls it map_statistics not sure how that works though

Sheikah45:
Reviews please?

migrations/V116__map_play_count_trigger.sql
```diff
@@ -0,0 +1,69 @@
...
+ALTER TABLE map
+    ADD COLUMN times_played int unsigned NOT NULL DEFAULT 0 COMMENT 'Number of times a map version has been played. Should be updated each time a row is inserted with a map_version.map.id in game_stats Values start at 0' AFTER reviews;
+
+-- one time update of table_map_features to get times_played to the right play_count
+UPDATE table_map_features,
+    (SELECT mapId, COUNT(DISTINCT id) AS map_play_count
+    FROM game_stats
+    GROUP BY mapId) as map_plays
+SET table_map_features.times_played = map_plays.map_play_count
+WHERE table_map_features.map_id = map_plays.mapId;
```
Brutus5000: Make it a view instead for compatibility while deprecating it (with a comment)

migrations/V116__map_play_count_trigger.sql
```diff
@@ -0,0 +1,69 @@
...
+WHERE map_version.id = map_plays.mapId;
+
+-- one time update of map_version to get times_played to the right play_count
+UPDATE map,
+    (SELECT map_version.map_id, COUNT(DISTINCT game_stats.id) AS map_play_count
+    FROM game_stats
+    INNER JOIN map_version on game_stats.mapId = map_version.id
+    GROUP BY map_version.map_id) as map_plays
+SET map.times_played = map_plays.map_play_count
+WHERE map.id = map_plays.map_id;
```
Brutus5000: These subselects are super costly from my experience. Better go for a join-based solution.

Brutus5000 (changes requested):

migrations/V116__map_play_count_trigger.sql
Sheikah45: Will do

migrations/V116__map_play_count_trigger.sql
Sheikah45: Would making it a view mean this query is run everytime someone wants to see the play count using the depreciated view which all clients currently do for any map in the vault? Would this be inefficient or am I missing something?

Sheikah45: Done although it looks like views can't have comments

Sheikah45:
Was there anything else we wanted to change here?

Brutus5000 (approved):

Brutus5000 (approved):

Issue #283, opened 2021-08-26
Store GOG user id


Brutus5000:
Similar to Steam Link if we allow GOG linking we need another column to store the user id.

PR #284, opened 2021-08-26
#283 Add column for gog.com username

Brutus5000:


Sheikah45:
Is the gog id really a string? Would be nice if steam and gog had the same type. But I am not sure what limits they put on their IDs

Brutus5000:
There is no point in parsing an id, because for looking up the games list we will (maybe even need to) user the username again.

Sheikah45:
Okay so this is storing the username then right. I was confused by the name

bukajsytlos (approved):

PR #286, opened 2021-09-11
Update readme to mention config file updates

BlackYps:
Simply trying to execute the migration will fail if the configuration files have changed in the meantime.
This adds explicit explanation how to fix this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,14 @@ You can now check the current version via running:
     select version from schema_version order by installed_rank desc limit 1
 
 ### Updating the database
-First you need to pull the faf-stack repo and checkout the branch with the db version you want. Then you need to update the configuration files.
+First you need to pull the faf-stack repo and checkout the branch with the db version you want. Now you can migrate to the latest version using:
+
+    docker-compose run --rm faf-db-migrations migrate
+
+The migration might fail if it depends on changes in the configuration files. In this case you can copy the template files again.
 
     cp -r config.template/* config
     cp .env.template .env
-
-Now you can migrate to the latest version using:
-
-    docker-compose run --rm faf-db-migrations migrate
 
 Updating the database is a one way street. Once you applied a migration, you cannot go back using a migration. If you want to do it manually you would need to undo the changes from the migrations and remove the version records in the table `schema_version`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,7 +28,12 @@ You can now check the current version via running:
     select version from schema_version order by installed_rank desc limit 1
 
 ### Updating the database
-First you need to pull the faf-stack repo and checkout the branch with the db version you want. Now you can migrate to the latest version using:
+First you need to pull the faf-stack repo and checkout the branch with the db version you want. Then you need to update the configuration files.
+
+    cp -r config.template/* config
+    cp .env.template .env
```
Sheikah45: If you are only updating the db you would only want to update the db config files where as this will overwrite all the config files so probably want to change this accordingly

README.md
```diff
@@ -28,7 +28,12 @@ You can now check the current version via running:
     select version from schema_version order by installed_rank desc limit 1
 
 ### Updating the database
-First you need to pull the faf-stack repo and checkout the branch with the db version you want. Now you can migrate to the latest version using:
+First you need to pull the faf-stack repo and checkout the branch with the db version you want. Then you need to update the configuration files.
```
Brutus5000: > Then you need to update the configuration files.

In **some** rare cases. I wouldn't recommend doing this everytime.
Or we make a suggestion on how to diff for changes.

BlackYps: So I just got unlucky? Then maybe the better workflow is to try the migration and if it fails then update the config files? Is it enough in this case to just update the db config file, or are there cases where other ones also need updates?

Brutus5000: I don't remember a single case where I broke the configuration for existing dbs.
Here is the history: https://github.com/FAForever/faf-stack/commits/develop/config.template/faf-db
Yes we added new users and database, but that does not break your existing database.

I'm not exactly sure which issue you encountered.

Brutus5000: Exemption: the recent Maria db migration. But I announced that in Zulip.

BlackYps: It happened when I tried to migrate from version 116 to 119

Brutus5000: The releases were in May and June. There were no config changes during that time :shrug:

BlackYps: No idea what happened exactly, but I think my config files have been quite old. No idea why that only caused issues so late instead of earlier.

Brutus5000 (approved):

PR #291, opened 2021-12-12
Enable 4v4 queues

BlackYps:
Once this is deployed the 4v4 queues should automatically get displayed in the client


Brutus5000 (approved):

Brutus5000:
Who gives me a heads up when to deploy?

BlackYps:
I guess Morax does. At least he should know when the map pool is ready.

BlackYps:
When this gets deployed it is absolutely critical that NEWBIE_TIME_BONUS in the server config gets set to 0. Otherwise everybody will get this bonus and the matching quality will be horrible. After some weeks we can set it back to the default value. I'm just not sure how we will remember to switch it back later.

PR #292, opened 2021-12-18
Ensure update time is updated when entity is changed

Sheikah45:
update_time colums were not set to actually reflect updates

### DIFF
--- a/migrations/V123__update_time_vault_items.sql
+++ b/migrations/V123__update_time_vault_items.sql
@@ -1,0 +1,11 @@
+ALTER table `map`
+    MODIFY COLUMN `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated';
+
+ALTER table `map_version`
+    MODIFY COLUMN `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated';
+
+ALTER table `mod`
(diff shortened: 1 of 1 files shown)


migrations/V123__update_time_vault_items.sql
```diff
@@ -0,0 +1,11 @@
...
+    MODIFY COLUMN `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated';
+
+ALTER table `map_version`
+    MODIFY COLUMN `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated';
+
+ALTER table `mod`
+    MODIFY COLUMN `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated';
+
+ALTER table `mod_version`
+    MODIFY COLUMN `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated';
```
norraxx, line 11: I'd use combination of two columns, created at, updated at.

norraxx (commented):
One proposition :-)
But other is good

migrations/V123__update_time_vault_items.sql
Sheikah45: create_time already exists. This is a fix for the update_time as before the on update was not set on these tables so it was only ever set at create time

Brutus5000 (approved):

PR #294, opened 2022-01-03
Add service links table

Sheikah45:


Askaholic:
This will need changes in the lobby server. Even for token login the lobby server still seems to be checking the database for steamid.

Sheikah45:
I thought the lobby server was checking steamid anymore for tokens? That was all on the user service. Because didnt we have that issue where for a short while the tokens could be used to log in without linking.

Askaholic:
I guess it doesn't enforce steam linking but it uses the presence of a steam id to determine if it can ignore the policy server result. If it can't find steam ids then people might get blocked from logging in by the policy server.

https://github.com/FAForever/server/blob/8902c127038383dcd5a1ad24324a2545ff5761ee/server/lobbyconnection.py#L537

Sheikah45:
ah gotcha, but regardless this change would just have that db call be on a separate table they are still linked by user id

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
+                         `user_id` mediumint(8) unsigned NOT NULL,
```
Brutus5000: Must be nullable for GDPR deletions.

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
...
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
+                         `service_id` varchar(100),
+                         `hashed_id` char(256) NOT NULL,
+                         `hash_salt` char(100) NOT NULL,
```
Brutus5000: This must be a global secret otherwise we can't check for existing hashes.

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
```
Brutus5000: Make this a uuid instead. Auto increment makes it ordered by time and makes it possible to narrow down for which registrations this link was based on near links before and after this one.

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
...
+                         `user_id` mediumint(8) unsigned NOT NULL,
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
+                         `service_id` varchar(100),
+                         `hashed_id` char(256) NOT NULL,
+                         `hash_salt` char(100) NOT NULL,
+                         `public` boolean,
+                         `create_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'When this entry was created.',
+                         `update_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'When this entry was updated',
```
Brutus5000: Delete these columns. Same reason as for id

Brutus5000 (changes requested):

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
+                         `user_id` mediumint(8) unsigned NOT NULL,
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
+                         `service_id` varchar(100),
+                         `hashed_id` char(256) NOT NULL,
```
Brutus5000: rename to `service_id_hash`

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
+                         `user_id` mediumint(8) unsigned NOT NULL,
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
+                         `service_id` varchar(100),
```
Brutus5000: Add a comment: "To be set to null if account is deleted"

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,30 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `id` mediumint(8) unsigned NOT NULL AUTO_INCREMENT,
+                         `user_id` mediumint(8) unsigned NOT NULL,
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
```
Brutus5000: Where do we define that this link confirms ownership? This information should not be hardcoded into the type.
I think on account deletion all links **not** related to ownership confirmation shall be deleted. So either each service link has a flag or we need a table for each type.

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,26 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `uuid` uuid unsigned NOT NULL,
+                         `user_id` mediumint(8) unsigned COMMENT "To be set to null if account is deleted",
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
+                         `service_id` varchar(100) COMMENT "To be set to null if account is deleted",
+                         `public` boolean,
+                         `ownership` boolean,
```
Brutus5000: Add a comment here:
`If true, this link confirms that the user owns FAForever. If true the link must never be deleted even if the linked account gets deleted. In this case we keep a dangling link so the id cannot be reused.`

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,26 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `uuid` uuid unsigned NOT NULL,
```
Brutus5000: uuid column should be named id.
also type uuid was just introduced with mariadb 10.7
until then use `id CHAR(36) CHARACTER SET ascii`

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,25 @@
...
+                         `type` enum('STEAM', 'GOG', 'DISCORD', 'PATREON') COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'The service that this service id links to.',
+                         `service_id` varchar(100) COMMENT 'To be set to null if account is deleted unless ownership is true',
+                         `public` boolean,
+                         `ownership` boolean COMMENT 'If true, this link confirms that the user owns FAForever. If true the link must never be deleted even if the linked account gets deleted. In this case we keep a dangling link so the id cannot be reused',
+                         PRIMARY KEY (id),
+                         UNIQUE KEY `unique_service_id` (`type`, `service_id`),
+                         CONSTRAINT `user_ibfk_1` FOREIGN KEY (`user_id`) REFERENCES `login` (`id`) ON DELETE RESTRICT ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci COMMENT='table representing the links to various services';
+
+# MIGRATION SCRIPT
```
Brutus5000: You need to create the ids yourself in the migration or we set the `id` default to `UUID() `

migrations/V124__create_links_table.sql
```diff
@@ -0,0 +1,25 @@
+DROP TABLE IF EXISTS `service_links`;
+
+CREATE TABLE `service_links` (
+                         `id` CHAR(36) CHARACTER SET ascii NOT NULL DEFAULT UUID(),
```
Brutus5000: After test I figured out that unfortunately UUID is not random at all (it's monotonic ascending). It's a UUID v1 so this implies the same problems as auto_increment.

We should remove the default and have it be populated by the application. Not sure how to fix the migration though.

Sheikah45: Is there some field we could sort by that would be more random?

Sheikah45: I fixed the migration script so that the uuids will be randomly ordered

Brutus5000 (approved):

PR #295, opened 2022-01-03
Add table representing the git commits for a version

Sheikah45:


Brutus5000 (changes requested):
If this is new logic why are we reusing the weird one table per featured mod thing?
Shouldn't we refer to the featuredMod table itself?

Sheikah45:
I wasn't sure if the table per featuredMod was old or new. I can definitely make it one table

Sheikah45:
Table for now

PR #297, opened 2022-01-22
Add licensing options on map/mod upload

Brutus5000:
License texts t.b.d.

Due to the nature of repeatable migrations we need V124 applied first on production, before further migrations can happen and nullability rules can be changed. Also map/mod-upload needs to be adapted before.


migrations/V127__vault_licenses.sql
```diff
@@ -0,0 +1,25 @@
...
+    revocable       boolean      not null comment 'Can the author revoke this license?',
+    redistributable boolean      not null comment 'Are other users (apart from FAF) allowed to redistribute this asset?',
+    modifiable      boolean      not null comment 'Are other users allowed to publish assets remixed, transformed, or built upon the material from this asset?',
+    constraint license_pk
+        primary key (id)
+)
+    comment 'Copyright licenses to be picked for uploaded content';
+
+alter table map
+    add license int null, -- nullability to be removed once all existing entries are migrated to undefined license
```
norraxx: Maybe we could place number `1` immediately? Turn on constraint check and then after migrations turn it on?

Brutus5000: This sounds like an idea. Start with default 1, but remove it right away again in the same migration, as in the long-term there is no guarantee that there is an id 1.

PR #298, opened 2022-02-13
Remove non-functioning featured mods

Sheikah45:


richi235:
I support this, we tried to play murder party yesterday and it cost us over 30 minutes of try and error until we had to realize it's broken anyway, a bit frustrating. Please merge quickly

Also the patch looks save to me as it only changes data: two database rows that might be displayed in the client (or not, if deleted) no critical code path or logic changes or anything involving dependencies. This is exactly how I would have done it too

richi235:
Maybe with some patch to remove the lua code for the mods later, but that might also be server side and another repo, I don't know your architecture very well

Brutus5000:
The preferred way is to offer regular mods offering the same functionality.
Featured mod cause a lot of problems in updating, patching and also are not compatible with other featured mods. (e.g. you can't really test a featured mod against fafdevelop, which is important to ensure your mod works in future FAF patches)

Sheikah45:
> Maybe with some patch to remove the lua code for the mods later, but that might also be server side and another repo, I don't know your architecture very well

There isn't any Lua code in the base game to support the mods. Really featured mods are supposed to be independent and supply all their own code. But that is not always the case which is why some break. This is the best way as it just removes them from usage by players.

Brutus5000 (approved):

PR #299, opened 2022-06-17
Add coturn servers table

Brutus5000:


Askaholic:
Is the lobby server going to need to implement this? Or will it be handled by the API?

Brutus5000:
API exposing the list to all clients

migrations/V124__coturn_servers.sql
```diff
@@ -0,0 +1,13 @@
+create table coturn_servers
+(
+    id            int(11) unsigned     not null auto_increment,
+    region        varchar(20)          not null comment 'Region of the server for simple user-based selection if latency checks don''t work',
+    host          varchar(255)         not null,
+    port          mediumint            not null,
```
Sheikah45: The current coturn server have only hosts and not ports so this should be nullable

Sheikah45 (changes requested):

migrations/V124__coturn_servers.sql
Sheikah45: Maybe the port should just be included in the host anyway?

Sheikah45: Also shouldn't the unique be host and port?

Sheikah45 (approved):

PR #305, opened 2023-03-19
Closes #304 Add average score virtual column to replay/maps/mods

### LINKED ISSUE #304: Add average score virtual column to replay/maps/mods

bukajsytlos:
Closes #304 Add average score virtual column to replay/maps/mods

### DIFF
--- a/migrations/V128__reviews_average_score.sql
+++ b/migrations/V128__reviews_average_score.sql
@@ -1,0 +1,30 @@
+alter table game_reviews_summary add (
+    average_score float as (case when reviews = 0 then null
+                                 else (score / reviews) end) virtual
+);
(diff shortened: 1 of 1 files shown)


Sheikah45 (commented):
Did you add in the case when just for cross db safety and clarity?

bukajsytlos:
deleting reviews in test data causes error division by zero. I am not sure why this is issue here and not in normal selection.
however it looks like it should throw error instead of returning null for newer versions (https://mariadb.com/kb/en/sql-mode/#error_for_division_by_zero). So to be safe, I made it like this

Sheikah45:
Sounds good

Sheikah45 (approved):

Brutus5000 (approved):

PR #312, opened 2023-06-15
Add GDPR deletion processing queues

Brutus5000:


migrations/V131__gdpr_delete_handling.sql
```diff
@@ -0,0 +1,15 @@
+create table account_deletion_requests
+(
+    id           varchar(36)                                                   null comment 'random uuid (as it''s also used for lookup of state by user)',
+    account_id   mediumint                                                     not null,
+    state        enum ('NEW', 'IN_PROCESS', 'FINISHED', 'ERROR') default 'NEW' not null,
+    full_delete  boolean                                                       not null comment 'full_delete=true means the whole account was deleted. This requires no games to be played.',
+    processed_by mediumint                                                     null comment 'Refers to the admin manually processing. Should be null if automatically processed',
+    comment      text                                                          null comment 'Allows manuell comments for manual processing',
```
Sheikah45:
```suggestion
    comment      text                                                          null comment 'Allows manual comments for manual processing',
```

migrations/V131__gdpr_delete_handling.sql
```diff
@@ -0,0 +1,15 @@
+create table account_deletion_requests
+(
+    id           varchar(36)                                                   null comment 'random uuid (as it''s also used for lookup of state by user)',
+    account_id   mediumint                                                     not null,
+    state        enum ('NEW', 'IN_PROCESS', 'FINISHED', 'ERROR') default 'NEW' not null,
+    full_delete  boolean                                                       not null comment 'full_delete=true means the whole account was deleted. This requires no games to be played.',
+    processed_by mediumint                                                     null comment 'Refers to the admin manually processing. Should be null if automatically processed',
```
Sheikah45, line 7: how is the id determined?

Brutus5000: account_id will be the id of the login id. but it can't have a foreign key if we want to delete that row

Sheikah45: Got it also forgot that was medium int too

migrations/V131__gdpr_delete_handling.sql
```diff
@@ -0,0 +1,15 @@
+create table account_deletion_requests
+(
+    id           varchar(36)                                                   null comment 'random uuid (as it''s also used for lookup of state by user)',
```
bukajsytlos, line 3: why not uuid data type?

Brutus5000: not sure how good the compatibility with raw sql, hibernate and python is

PR #313, opened 2023-06-15
Add handling of accepted ToS on account level

Brutus5000:
The idea here is the following: ON each login of the user service we check whether the user has accepted the latest ToS, otherwise he can't login. This is a known smooth way to gradually transition to newer ToS versions.

### DIFF
--- a/migrations/V132__add_tos.sql
+++ b/migrations/V132__add_tos.sql
@@ -1,0 +1,16 @@
+create table terms_of_service
+(
+    version    smallint auto_increment,
+    valid_from timestamp not null comment 'The start date when this ToS version is applied. May point into the future to prepare ToS changes.',
+    content    text      not null comment 'Markdown formatted text',
+    constraint terms_of_service_pk
+        primary key (version)
+);
+
+create index terms_of_service_valid_from_index
+    on terms_of_service (valid_from);
+
+alter table login
+    add accepted_tos smallint null comment 'Point to the last ToS the user has explicitely accepted',
+    add constraint login_terms_of_service_version_fk
+        foreign key (accepted_tos) references terms_of_service (version);


Brutus5000:
I'm actually wondering if it makes more sense to have a link to the ToS instead, e.g. on a Github pages site with each releases. Then we could easily show the diffs between versions and have less mismatch between website and database (website could also point to the Github page instead).

Sheikah45:
That depends, a website link could easily get changed or lost. Making it confusing to rediscover. Maybe if there was some way to serve directly from the database

Brutus5000:
Github pages with releases and history? Isn't getting lost. But the formatting only applies to the latest version ofc.

migrations/V132__add_tos.sql
```diff
@@ -0,0 +1,16 @@
+create table terms_of_service
+(
+    version    smallint auto_increment,
+    valid_from timestamp not null comment 'The start date when this ToS version is applied. May point into the future to prepare ToS changes.',
```
bukajsytlos, line 4: how would one find current version? wouldn't it be better to have also valid_to?

Brutus5000: order by version desc limit 1
having valid_to and valid_from leads to potential inconsistencies as valid_from must be the valid_to the predecessor which cannot be enforced by mysql

bukajsytlos: but if you insert planned one?

Brutus5000: damn

Brutus5000: on the other hand... even if I add valid_to I need to check valid_from <= CURRENT_TIMESTAMP

bukajsytlos: ah right, so `valid_from <= CURRENT_TIMESTAMP order by version desc limit 1`

Brutus5000: Even more correct would be
```
WHERE valid_from <= CURRENT_TIMESTAMP
ORDER BY valid_from DESC
LIMIT 1
```

in case someone messes up with inserting

bukajsytlos (approved):
lgtm

PR #315, opened 2024-01-12
Don't allow players to friend or foe themselves

Askaholic:
It doesn't make sense to allow friend/foe relationships to ones self. Are we supposed to hide our own chat messages from ourselves if we are our own foe? Are we supposed to be disallowed from playing a game with ourselves?

I think the client already isn't supposed to allow you to do this, but I can see in the server logs that sometimes this is attempted. I would be curious to know the number of self referential rows there are on prod.

```
select count(*) from friends_and_foes where user_id=subject_id;
```


Brutus5000:
We have 43 self-hating users.

Brutus5000 (approved):

Issue #317, opened 2024-04-27
veto_count fields to matchmaker_queue


K-ETFreeman:
need to add veto_count fields to matchmaker_queue
they will store the max count of vetoes which can be applied by player to each queue
(this is part of the [veto system development](https://github.com/FAForever/downlords-faf-client/issues/2319))

PR #318, opened 2024-04-27
Veto_count fields for matchmaker_queue

K-ETFreeman:
Solves [#317](https://github.com/FAForever/db/issues/317) 
Atleast, i think so. But i have very slight idea what am i doing so please check it twice before merging.

### DIFF
--- a/migrations/R__070_matchmaker_queue.sql
+++ b/migrations/R__070_matchmaker_queue.sql
@@ -4,12 +4,12 @@
 -- If you want to edit an existing or add a new matchmaker queue, just append it to the SQL command and run flyway migrate.
 -- Matchmaker queues are handled dynamically. If they are enabled they will show up. (The map pool is managed by privileged users.)
 
-INSERT INTO `matchmaker_queue` (id, technical_name, featured_mod_id, leaderboard_id, name_key, team_size, params, enabled) VALUES
-(1,'ladder1v1',0,2,'matchmaker_queue.ladder1v1',1,NULL,1),
-(2,'tmm2v2',0,3,'matchmaker_queue.tmm2v2.name',2,NULL,1),
-(3,'tmm4v4_full_share',0,4,'matchmaker_queue.tmm4v4_full_share.name',4,NULL,1),
-(4,'tmm4v4_share_until_death',0,5,'matchmaker_queue.tmm4v4_share_until_death.name',4,'{"GameOptions":{"Share":"ShareUntilDeath"}}',0),
-(5,'tmm3v3',0,6,'matchmaker_queue.tmm3v3.name',3,NULL,1)
+INSERT INTO `matchmaker_queue` (id, technical_name, featured_mod_id, leaderboard_id, name_key, team_size, params, enabled, veto_tokens_count) VALUES
+(1,'ladder1v1',0,2,'matchmaker_queue.ladder1v1',1,NULL,1,0),
+(2,'tmm2v2',0,3,'matchmaker_queue.tmm2v2.name',2,NULL,1,0),
+(3,'tmm4v4_full_share',0,4,'matchmaker_queue.tmm4v4_full_share.name',4,NULL,1,0),
+(4,'tmm4v4_share_until_death',0,5,'matchmaker_queue.tmm4v4_share_until_death.name',4,'{"GameOptions":{"Share":"ShareUntilDeath"}}',0,0),
+(5,'tmm3v3',0,6,'matchmaker_queue.tmm3v3.name',3,NULL,1,0)
 -- add new row above this comment (don't forget to append the comma to the previous one)
 ON DUPLICATE KEY UPDATE
     technical_name=VALUES(technical_name),
@@ -17,4 +17,5 @@ ON DUPLICATE KEY UPDATE
     leaderboard_id=VALUES(leaderboard_id),
     name_key=VALUES(name_key),
     team_size=VALUES(team_size),
-    enabled=VALUES(enabled);
+    enabled=VALUES(enabled),
+    veto_tokens_count=COALESCE(veto_tokens_count, VALUES(veto_tokens_count));
--- a/migrations/V135__add_veto_count_to_matchmaker_queues.sql
+++ b/migrations/V135__add_veto_count_to_matchmaker_queues.sql
@@ -1,2 +1,0 @@
-ALTER TABLE matchmaker_queue
-  ADD COLUMN veto_count TINYINT UNSIGNED NOT NULL DEFAULT 0;
--- a/migrations/V135__add_veto_count_to_matchmaker_queues.sql
+++ b/migrations/V135__add_veto_count_to_matchmaker_queues.sql
@@ -1,0 +1,2 @@
+ALTER TABLE matchmaker_queue
+  ADD COLUMN veto_count TINYINT UNSIGNED NOT NULL DEFAULT 0;
--- a/migrations/V135__add_veto_tokens_count_to_matchmaker_queues.sql
+++ b/migrations/V135__add_veto_tokens_count_to_matchmaker_queues.sql
@@ -1,0 +1,2 @@
+ALTER TABLE matchmaker_queue
+  ADD COLUMN veto_tokens_count TINYINT UNSIGNED NOT NULL DEFAULT 0;


Brutus5000 (commented):
The change is ok. 
To actually set the values for the matchmakers, the repeatable migration here needs to be adjusted then too:
https://github.com/FAForever/db/blob/develop/migrations/R__070_matchmaker_queue.sql

I won't merge until I see decent corresponding activity in the lobby server repo though.

K-ETFreeman:
@Brutus5000 Hi
We (me & sheikah) were discussing the possible options for the veto system, and the veto_count field can be actually added to any of:
matchmaker_queue
matchmaker_queue_map_pool
map_pool
collections (tables). The idea is to add some api endpoint which will let matchmaker team members to modify the veto_count value for existing documents (rows).
Want to hear your thoughts about that, what is the better way you think

Brutus5000:
That depends.`matchmaker_queue` is the right place if it's a global setting per matchmaker. If you want different veto-blocks settings per rating range, `map_pool`  is the right place. `matchmaker_queue_map_pool`  is just for getting the right map pool per matchmaker.

I can make the API change, this is not a big deal. Just adding a field with already used permissions. The thing that takes the longest from my experience is the lobby server.

BlackYps:
Shouldn't we use `matchmaker_queue_map_pool`? That one seems to be tasked with additional settings of a map pool. Exposing this table on the API would also make it easier for the matchmaker team to change the rating ranges.
But to be honest I don't really understand why the `map_pool` table exists at all, so maybe my understanding is wrong here.

K-ETFreeman:
> That depends.`matchmaker_queue` is the right place if it's a global setting per matchmaker. If you want different veto-blocks settings per rating range, `map_pool` is the right place. `matchmaker_queue_map_pool` is just for getting the right map pool per matchmaker.
> 
> I can make the API change, this is not a big deal. Just adding a field with already used permissions. The thing that takes the longest from my experience is the lobby server.

Ah, right. We definitely want it to set per matchmaker. 
(I forgot a bit that map_pool is only 1 rating bracket, not actual map pool for a user)
So, better will be to set it into matchmaker_queue
However, the idea is modified a bit, matchmaker team now wants more flexibility with "veto tokens", so we can in theory add ability to "half veto" some map, if we want. The details are not set clearly yet, but anyway i rename the field from **veto_count** to **veto_tokens_count**
Also i modified the repeated migration file, i hope this was done correctly. We dont want for veto_tokens_count being resetted each migration, so i used COALESCE for it, maybe it works now?

Sheikah45:
Do we want the veto count to be at the matchmaker level though? 

If the goal is to set the number of tokens so that not all maps can be vetoed I would think it would be better to set it at the pool level since each pool can have a different number of maps. So setting it at the matchmaker level seems a little restrictive.

K-ETFreeman:
> Do we want the veto count to be at the matchmaker level though?
> 
> If the goal is to set the number of tokens so that not all maps can be vetoed I would think it would be better to set it at the pool level since each pool can have a different number of maps. So setting it at the matchmaker level seems a little restrictive.

Each pool is not actual pool, its 1 bracket, as i understand
We dont want to have separate vetoes for each bracket isnt it?

Brutus5000:
> Shouldn't we use `matchmaker_queue_map_pool`? That one seems to be tasked with additional settings of a map pool. Exposing this table on the API would also make it easier for the matchmaker team to change the rating ranges. But to be honest I don't really understand why the `map_pool` table exists at all, so maybe my understanding is wrong here.

The map pool is the entity that holds the maps.
Technically a map pool has no knowledge about the existence of matchmaker queues. It could be a map pool for a tournament or whatever.

matchmaker_queue_map_pool is the m:n relationship table between matchmaker queues and map pools. But it needs attributes to query the right map pool: the rating range. These attributes are matchmaker related.

The amount of allowed vetos is a map pool feature, and not relevant for picking the correct map pool. Thus it should not go into `matchmaker_queue_map_pool`.

Is this system overengineered? Maybe. Is it based on my input? Maybe. I don't remember who came up with this.
Nevertheless I would say this is high quality data design. No coupling where there shouldn't be any. It may cause a little more work on ui side, but we remain flexible.

We can still just add the flag to the api on the map pool entity without any issues.

BlackYps:
I get it now. OK, so if the map pool could theoretically also be used for tournaments or similar stuff then I agree that we should add the veto column to that table. We will probably use the same count for all pools in a matchmaking queue but I can maybe see a use case with different pool sizes in a single queue that warrant different veto counts. But even if we don't do that, if we ever use these pools for tournaments then it would be nice if we already have the basics set up to also support vetoing there. So it isn't necessarily coupled to a matchmaking queue.

K-ETFreeman:
> I get it now. OK, so if the map pool could theoretically also be used for tournaments or similar stuff then I agree that we should add the veto column to that table. We will probably use the same count for all pools in a matchmaking queue but I can maybe see a use case with different pool sizes in a single queue that warrant different veto counts. But even if we don't do that, if we ever use these pools for tournaments then it would be nice if we already have the basics set up to also support vetoing there. So it isn't necessarily coupled to a matchmaking queue.

Having veto count setted in map pools would make sense if map pool was entire map pool for a user, but its only 1 rating bracket
If we set it into that, client would take all the brackets, and then what? take first bracket veto count? take last bracket veto count? take random bracket veto count? just hoping they all the same? but what if they are not, for some reason?
it doesnt seem any good
tournaments are really far from using map pools and veto sytem for now, it will make sense to care about them only if we want to make full-inside-client tournament generator system, or something like that. 

However, if you really want to make it right, we need to create entirely new entity in the database "actual_map_pool" which will contain the complete set of all brackets for some queue or for some tournament. Then yes, it would be good to set veto_tokens_count there

Alternative would be to just modify matchmaker_queue so it can also be used in a tournament somehow. So, when we want to host another tournament, we create new matchmaker_queue and setting another veto_count for it. The name should be changed then and i am not sure how it can be modified to be both tournament and queue entity 

But do we really need to do that complicated stuff just for tournament system idea maybe some day being real?

Sheikah45:
To me it makes more sense to set the vetoes at the map pool level. This is because each pool or bracket can have a different number of maps in it. And then we would give players a veto count based off the number of maps in the pool. Otherwise we end up in a situation where low rating players are able to veto every map because we give enough vetoes so that high rating players can veto some fraction of their maps.

K-ETFreeman:
> To me it makes more sense to set the vetoes at the map pool level. This is because each pool or bracket can have a different number of maps in it. And then we would give players a veto count based off the number of maps in the pool. Otherwise we end up in a situation where low rating players are able to veto every map because we give enough vetoes so that high rating players can veto some fraction of their maps.

But players are not hardly connected to their bracket. They still can get maps from lower or higher bracket. So they need to have ability to veto map from lower or higher bracket. And it would be really strange if we using player bracket as source of veto tokens

Imagine really large bracket with a lot of maps/vetoes near really small bracket with just few maps/vetoes. Then two players near edge of the bracket range with 2 rating points difference. They will have the same set of maps basically, but one of them will have about no veto tokens, while other will have a lot of veto tokens. Bad.

And if you want to suppose separate vetoes for each bracket, it will overcomplicate the UI and also has problems with "same map in multiple brackets" case

K-ETFreeman:
However it maybe can work if we make veto_tokens being calculated as some function of (player rating, brackets)
it is hard tho because we need to understand probability of brackets really well and this is not easy as we already discussed

Sheikah45:
Or you just allow players to veto maps on a bracket pool basis rather than at the matchmaker level.

Then when the server picks the pool for the game it uses the players vetos for that pool.

This makes more sense as otherwise you would also have players using vetos when they aren't really going to get that map.

Sheikah45:
And the nature of the complicated UI or options is just because the actual system itself is complicated.

I think trying to pretend or simplify it will just lead to headaches down the road.

I suspect players should just be able to select vetoes for all brackets regardless of their rating since players can get maps from any bracket.

BlackYps:
ETFreeman raised a good point in discord: The amount of total vetos depends on the number of players in a matchmaker queue. So it should not be a property of just the map pool.
To decide what a sensible amount of vetos is per player we need the combined info from the matchmaker queue and the map pool. So the fitting table would be `matchmaker_queue_map_pool`.

Brutus5000:
`matchmaker_queue_map_pool` has a one-to-one relationship with map pools. If it doesn't fit into the map pool it will most likely and conceptually not fit in the m:n table either.

But reading this convo I'm a bit lost on how the map veto is supposed to be calculated and it would be much better if someone can come up with a decision table or a formular how to get the number of veto counts.
To be honest to me it feels like we are scratching something like (team_size * sqrt(size of selected map pool)) and then it doesn't need any additional stored attributes at all.

BlackYps:
> `matchmaker_queue_map_pool` has a one-to-one relationship with map pools. If it doesn't fit into the map pool it will most likely and conceptually not fit in the m:n table either.

I don't see the conceptual problem? Have a look at the following example data:
| matchmaker_queue_id | map_pool_id | min_rating | max_rating | veto_count |
|  3                                   |  1                     |  500             |  1000           |  4                  |
|  4                                   |  1                     |  500             |  1000           |  2                  |

As you can see we can now specify a different veto count for the same map pool when it is used by different queues


> But reading this convo I'm a bit lost on how the map veto is supposed to be calculated and it would be much better if someone can come up with a decision table or a formular how to get the number of veto counts.
> To be honest to me it feels like we are scratching something like (team_size * sqrt(size of selected map pool)) and then it doesn't need any additional stored attributes at all.

That's probably because we are still in the phase where we discover that the originally envisioned approach won't work and we need to come up with a new concept. When we have the concept it will be clear what changes exactly we need in the db.

Brutus5000:
> I don't see the conceptual problem? Have a look at the following example data:
> | matchmaker_queue_id | map_pool_id | min_rating | max_rating | veto_count |
> | 3 | 1 | 500 | 1000 | 4 |
> | 4 | 1 | 500 | 1000 | 2 |

Ok I see now what you mean. You anticipate the case that a map pool could be used in multiple queues which is not the case right now. But yeah, that might make sense depending on the formula to be applied.

Brutus5000:
Closed due to inactivity. I have not seen any discussions to bring this forward on any channels, especially regarding calculation the map veto threshold.

Can be reopened if anything changes on that behalf.

Issue #319, opened 2024-10-12
Add new columns to matchmaker_queue_map_pool table


K-ETFreeman:
Need to add 3 columns, veto_tokens_per_player, max_tokens_per_map and minimal_maps_allowed to matchmaker_queue_map_pool table
This is a part of [veto system development](https://github.com/FAForever/downlords-faf-client/issues/2319) ([description of veto system functionality](url))

PR #320, opened 2024-10-12
#319 adding 3 columns to matchmaker_queue_map_pool table

K-ETFreeman:
solves #319

### DIFF
--- a/migrations/V135__add_veto_fields_to_matchmaker_queue_map_pool
+++ b/migrations/V135__add_veto_fields_to_matchmaker_queue_map_pool
@@ -1,0 +1,4 @@
+ALTER TABLE matchmaker_queue_map_pool
+  ADD COLUMN veto_tokens_per_player TINYINT UNSIGNED DEFAULT 0 NOT NULL,
+  ADD COLUMN max_tokens_per_map TINYINT UNSIGNED DEFAULT 0 NOT NULL,
+  ADD COLUMN minimal_maps_allowed FLOAT DEFAULT 1 NOT NULL CHECK (minimal_maps_allowed >= 1);
--- a/migrations/V135__add_veto_fields_to_matchmaker_queue_map_pool.sql
+++ b/migrations/V135__add_veto_fields_to_matchmaker_queue_map_pool.sql
@@ -1,4 +1,4 @@
 ALTER TABLE matchmaker_queue_map_pool
   ADD COLUMN veto_tokens_per_player TINYINT UNSIGNED DEFAULT 0 NOT NULL,
   ADD COLUMN max_tokens_per_map TINYINT UNSIGNED DEFAULT 0 NOT NULL,
-  ADD COLUMN minimal_maps_allowed FLOAT DEFAULT 1 NOT NULL CHECK (minimal_maps_allowed >= 1);
+  ADD COLUMN minimum_maps_after_veto FLOAT DEFAULT 1 NOT NULL CHECK (minimum_maps_after_veto > 0);


migrations/V135__add_veto_fields_to_matchmaker_queue_map_pool
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE matchmaker_queue_map_pool
+  ADD COLUMN veto_tokens_per_player TINYINT UNSIGNED DEFAULT 0 NOT NULL,
+  ADD COLUMN max_tokens_per_map TINYINT UNSIGNED DEFAULT 0 NOT NULL,
+  ADD COLUMN minimal_maps_allowed FLOAT DEFAULT 1 NOT NULL CHECK (minimal_maps_allowed >= 1);
```
BlackYps: I would name `minimal_maps_allowed` differently. It is a restriction, not an allowance. I get what it is supposed to mean, but it is a bit confusing. Maybe `minimal_maps_left`?

Brutus5000: I don't understand either of the names. What is the effect of this value?

K-ETFreeman: > I don't understand either of the names. What is the effect of this value?

its the minimal maps count which should left after all vetoes applied

K-ETFreeman: system will nerf veto tokens if necessary until this is true

K-ETFreeman: about the name i dont really care, but PRs to the api and mod client should also be changed accordingly, if you change it

K-ETFreeman: > > I don't understand either of the names. What is the effect of this value?
> 
> its the minimal maps count which should left after all vetoes applied

predicting the question "why it is float", 1 map here is a map with weight of 1. 0.5 map is a map with weight of 0.5, etc.
the more weight the more probability

K-ETFreeman: also if you want you can replace to CHECK (minimal_maps_allowed > 0), it will still technically work with 0.1 values, but not with 0

K-ETFreeman: > I would name `minimal_maps_allowed` differently. It is a restriction, not an allowance. I get what it is supposed to mean, but it is a bit confusing. Maybe `minimal_maps_left`?

and about "not an allowance"
image
translator is wrong "cards" with "maps" here, but you got the idea
so in general the name is fine

Brutus5000: I'm a big fan of self-describing names.
And the only one I come up with really explaining what it is would be `minimum_maps_after_veto`.

The `_after_veto` may look superfluous but it's not because leaving it out can lead to wrong understanding due to the dynamic amount of maps in a pool anyway (it could refer to the amount of maps that need to be assigned to that pool).

BlackYps: I like `minimum_maps_after_veto`. It's a descriptive name.

K-ETFreeman: ok i will set "minimum_maps_after_veto" and will update other repos / PRs accordingly

K-ETFreeman: i want to set it like CHECK (minimum_maps_after_veto > 0)
but this can lead to questions like "what is 0.5 maps after veto means"
should i rename it to "minimal_sum_of_map_weights_after_veto" instead?

K-ETFreeman: that would be more accurate description of the variable

K-ETFreeman: @BlackYps @Brutus5000

K-ETFreeman:
renamed the column

PR #327, opened 2025-09-23
Add ip to game join log + indices

Brutus5000:


Sheikah45:
Why not just an index on IP by itself? The double index seems aggressive

Brutus5000:
Because its 2 different lookup pattern. Look for all players in a suspicious ip range. And then look for all ips of a suspicious player

Sheikah45:
yeah but if you are just looking for all in both of those sets then two independent indicies would be enough

bukajsytlos:
afaik only one index is used at a time

